PR #10, opened 2019-03-15
Post equal height

prajaktagadhave123:
* Improvement: Post Carousel - Added equal height option.


src/blocks/post/post-carousel/edit.js
```diff
@@ -71,14 +71,24 @@ class UAGBPostCarousel extends Component {
 	}
 
 	componentDidMount() {
-
+		var id = this.props.clientId
```
vrundakansara: Why this extra variable?

classes/class-uagb-init-blocks.php
```diff
@@ -178,6 +178,15 @@ function block_assets() {
 			);
 		}
 
+		// Carousel js.
+		wp_enqueue_script(
```
vrundakansara: Enqueue this conditionally.

PR #65, opened 2019-05-27
Fix the preview and the frontend rendering of the quote and author styling options for the blockquote block.

kmgalanakis:
There have been issues with the preview/frontend rendering of the styling options for the blockquote block due to a wrong class name.

Source files should be OK with this PR as well as the bundled sources.


vrundakansara:
Hey @kmgalanakis 

Thank you for the PR. I have tested this and it works great! We appreciate your efforts.

Issue #88, opened 2019-07-24
Notice displaying Social Share block


bobbingwide:
Problem occurs because no background colour has been defined for the icons.

```
Notice: Undefined index: icon_bg_hover_color in /home/bworguk/public_html/wp-content/plugins/ultimate-addons-for-gutenberg/classes/class-uagb-block-helper.php on line 1742
```

See https://blocks.wp-a2z.org/block/social-share-uagb-social-share/

I raised this as a support ticket just a few moments ago. Then I matched bsf.io to brainstormforce, put two and two together and answered part of the ticket myself. 

Note: Problem no longer occurs on the above URL as I have applied the fix below.

bobbingwide:
Adding a couple of lines to `classes/class-uagb-block-helper.php` resolves the issue. Around line 1705
```
$social['icon_color'] = ( isset( $social['icon_color'] ) ) ? $social['icon_color'] : '';
$social['icon_bg_color'] = ( isset( $social['icon_bg_color'] ) ) ? $social['icon_bg_color'] : '';
$social['icon_hover_color'] = ( isset( $social['icon_hover_color'] ) ) ? $social['icon_hover_color'] : '';
$social['icon_bg_hover_color'] = ( isset( $social['icon_bg_hover_color'] ) ) ? $social['icon_bg_hover_color'] : '';
```

bobbingwide:
I've created PR #89

vrundakansara:
This has been merged and will be released in the upcoming version.

PR #89, opened 2019-07-25
Issue #88 - Cater for user clearing an icon's background colours

bobbingwide:
Need to set values for `icon_bg_color` and `icon_bg_hover_color` for when the user clicked on Clear


vrundakansara:
@bobbingwide - Thank you for the PR.

PR #107, opened 2019-09-23
Add support for EditorsKit navigator toolbar

phpbits:
Hi,

I've just recently added Block Navigation toolbar on my [EditorsKit](https://wordpress.org/plugins/block-options/) plugin. I think this will be of great help on user navigation when both plugins are activated. The toolbar will only be shown on blocks with multiple inner blocks capability.

- [x] Columns
- [x] Columns

Here's how it works:
Screen Capture on 2019-09-23 at 14-38-38

Let me know your thoughts.

Thanks,
Jeffrey


vrundakansara:
Thank you @phpbits . I will review and merge this.

PR #122, opened 2019-10-31
New filter

pixelated-au:
I created this filter-call to conditionally show the author depending on whether it's a local author or a link to another blog. In the case of the latter, we're not showing the author details. Therefore, I'm using the filter to set the 'displayPostAuthor' to false.
Hopefully, this is useful to you!


vrundakansara:
Thank you for the PR @pixelated-au .

PR #127, opened 2019-11-04
Implementing AdaptiveHeight for testimonial blocks

cheey2003:
Activate the AdaptiveHeight option for Slick Carousel so that the slide heights adapt to contents of different length


vrundakansara:
Hey @cheey2003 ,

This will change the designs for existing users. We will add such an option from the backend. Meanwhile, you can use this filter to achieve this - 

```php
add_filter( 'uagb_testimonials_slick_options', function( $attr, $id ) {
	$attr['adaptiveHeight'] = true;
	return $attr;
}, 10, 2 );
```
I hope that helps.

Regards,
Vrunda Kansara

cheey2003:
Thank you very much for the code snippet!

On Tue, Nov 5, 2019 at 4:51 PM, Vrunda Kansara <notifications@github.com> wrote:
Hey @cheey2003 [https://github.com/cheey2003] ,

This will change the designs for existing users. We will add such an option from the backend. Meanwhile, you can use this filter to achieve this -

add_filter( 'uagb_testimonials_slick_options', function( $attr, $id ) {$attr['adaptiveHeight'] =true;return$attr;}, 10, 2 );

I hope that helps.

Regards,
Vrunda Kansara

PR #158, opened 2020-03-02
v1.14.8

vrundakansara:
* Improvement: Inline CSS generation markup as per w3c standards.
* Improvement: Hardened the security of the plugin.
* Improvement: Compatibility with Toolset Views.
* Fix: Table of Contents - Scroll to Top issue in some scenarios.
* Fix: Font Families not applying if it's name contains a space.
* Fix: Undefined constant FS_CHMOD_FILE issue.
* Fix: WP Rocket - "javascript defered without safe mode jQuery" compatibility.


classes/class-uagb-init-blocks.php
```diff
@@ -48,6 +48,45 @@ public function __construct() {
...
+		add_action( 'wp_ajax_nopriv_uagb_cf7_shortcode', array( $this, 'cf7_shortcode' ) );
+	}
+
+	/**
+	 * Renders the Gravity Form shortcode.
+	 *
+	 * @since 1.12.0
+	 */
+	public function gf_shortcode() { 	// @codingStandardsIgnoreStart
+		$id = intval($_POST['formId']);
```
RajkiranBagal: Ignore PHPCS using inline comments.

vrundakansara: This is to ignore the camelCase variable - `formId`. This variable is not controlled by UAG.

PR #173, opened 2020-03-30
Fix: Border radius not working for post carousel

AbhijitNage123:
- remove border CSS from the slick.css which conflicting the border-radius setting.


assets/css/slick.css
```diff
@@ -146,7 +146,7 @@
 
 .uagb-slick-carousel button.slick-arrow {
 	padding: 7px;
-	border-radius:0;
+	/* border-radius:0; */
```
vrundakansara: Why leave the code commented if the code is not needed?
This is a common CSS for Testimonial and Post Carousel.
Make sure you test both thoroughly.

PR #175, opened 2020-03-31
Fix: the issue of reset button alignment in section

AbhijitNage123:
- Added display flex CSS to display reset button inline.


vrundakansara:
Add in changelog too. @AbhijitNage123

AbhijitNage123:
Done with updating changelog

PR #183, opened 2020-04-06
Percentage btn not clickable

AbhijitNage123:
-Added css to clickable percentage btn in ad column.


RajkiranBagal:
@AbhijitNage123  Update changelog

PR #412, opened 2020-11-04
Improvement: faq.js is not needed in the grid layout

AbhijitNage123:
### Description

Trello - https://trello.com/c/9d6p9eiC/110-faqjs-is-not-needed-in-the-grid-layout

### Types of changes

- Improvement

### How has this been tested?

- Used faq grid layout and check is it loading `uagb-faq-js` .
- Used some other block-like timeline, testimonials and inline notice. Is this block working when the faq layout is the grid?
- Use both faqs as grid and accordion and check accordion is working or not.

### Checklist:
- [x] My code is tested

### DIFF
--- a/classes/class-uagb-helper.php
+++ b/classes/class-uagb-helper.php
@@ -205,11 +205,11 @@ if ( ! class_exists( 'UAGB_Helper' ) ) {
 
 				foreach ( $js_assets as $asset_handle => $val ) {
 					// Scripts.
-					if( $val == 'uagb-faq-js'){
-						if( self::$uag_faq_layout ){
+					if ( $val === 'uagb-faq-js' ) {
+						if ( self::$uag_faq_layout ) {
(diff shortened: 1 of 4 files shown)


classes/class-uagb-helper.php
```diff
@@ -197,7 +197,24 @@ public function block_assets() {
 
 				foreach ( $js_assets as $asset_handle => $val ) {
 					// Scripts.
-					wp_enqueue_script( $val );
+					if( $val === 'uagb-faq-js' ){
```
vrundakansara, line 200: I think this will not be optimal. We need to implement the logic the same as we have done for `$uag_flag`.
You need to set a flag when the Grid layout is used.

Fetching the posts again and parsing them again will be unnecessary extra work.

AbhijitNage123: Done

vrundakansara (approved):
Looks good to me. 👍🏼

PR #505, opened 2021-03-08
Fix: Block Conditions - Conflicting with the WPForms.

AbhijitNage123:
### Description

Trello - https://trello.com/c/qGR6uGBc/209-fix-block-conditions-conflicting-the-wpforms

### Types of changes

### How has this been tested?

- Create a simple WPForm.
- Add Form to the page using WPForms block and check if an error occurred or not.

### Checklist:
- [x] My code is tested


RajkiranBagal (approved):

RajkiranBagal:
@AbhijitNage123 Just confirm the Changelog once.

src/blocks/extensions/attributes.js
```diff
@@ -1,40 +1,42 @@
...
-			UAGLoggedIn: {
-				type: "boolean",
-		        default: false
-			},
-			UAGLoggedOut: {
-				type: "boolean",
-		        default: false
-			},
-		} );		
+	if( settings.category === 'uagb' ){
```
vrundakansara, line 3: Does this mean we will have these visibility options just for UAG blocks?

AbhijitNage123: Yes, We only adding block attributes when the category is a match.

vrundakansara: This is not the expected behavior @AbhijitNage123 
We want it for all blocks.

Can you elaborate on what the conflict was?

AbhijitNage123: After drag and drop the WPForms on-page. It shows the `Error loading block: Invalid parameter(s): attributes`. And it seems that we are adding attributes for all the blocks. and due to which the attribute is added in the ajax URL - https://share.getcloudapp.com/qGuXWPxA because that block is enabled to render it. @vrundakansara

vrundakansara: We need to fix that issue then. See why these parameters are passed and how we can avoid them.

AbhijitNage123: Adding Responsive CSS through javascript in order to work for all blocks. And exclude the WPForms from the conditional block.

classes/class-uagb-block-helper.php
```diff
@@ -5518,5 +5518,18 @@ public static function get_lottie_css( $attr, $id ) {
 
 			return UAGB_Helper::generate_all_css( $combined_selectors, '.uagb-block-' . $id );
 		}
+
+		/**
+		 * Get Condition block CSS.
+		 *
+		 * @since x.x.x
+		 * @param array  $attr The block attributes.
+		 * @param string $id The selector ID.
```
vrundakansara, line 5527: Remove these two params since we don't have any function parameters.

vrundakansara (changes requested):

vrundakansara (approved):

PR #507, opened 2021-03-09
Fix: Block Conditions - Visible only to UAGB Block.

AbhijitNage123:
### Description

Trello - https://trello.com/c/2L8KlvMF/210-display-block-condition-only-to-uag-block

### Types of changes

### How has this been tested?

- Drag and drop UAG block and check display condition is visible in all blocks or not.
- Drag and drop any block like core or any competitor block and our display condition should not visible in that block.
- Also check if Responsive visibility working properly for a block.

### Checklist:
- [x] My code is tested


classes/class-uagb-config.php
```diff
@@ -2666,15 +2666,15 @@ public static function get_block_attributes() {
 						),
 					),
 					'uagb/condition-block'        => array(
-						'slug'        => '',
+						'slug'        => 'uagb',
```
RajkiranBagal, line 2669: Where is this slug used?

RajkiranBagal (changes requested):

RajkiranBagal (approved):

PR #650, opened 2021-05-05
Check upload dir permission and update asset generation option in fresh setup

sandeshjangam:
### Description
Check upload dir permission and enabled asset file generation option in fresh setup

### Types of changes
Non-breaking change

### How has this been tested?
Install UAG plugin in fresh setup.
File generation should be enabled if the `wp-upload` folder is writable

### Checklist:
- [x] My code is tested

### DIFF
--- a/classes/class-uagb-update.php
+++ b/classes/class-uagb-update.php
@@ -40,7 +40,7 @@ if ( ! class_exists( 'UAGB_Update' ) ) :
 		 *  Constructor
 		 */
 		public function __construct() {
-			add_action( 'admin_init', __CLASS__ . '::init' );
+			add_action( 'admin_init', array( $this, 'init' ) );
 		}
 
 		/**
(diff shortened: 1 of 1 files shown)


classes/class-uagb-update.php
```diff
@@ -49,31 +49,53 @@ public function __construct() {
 		 * @since 1.13.4
 		 * @return void
 		 */
-		public static function init() {
-
-			do_action( 'uagb_update_before' );
+		public function init() {
```
RajkiranBagal, line 52: Jut curious why are we replacing a static function with a  object function?

sandeshjangam: We do not use that function without an instance, so, not necessary here.

Ideally, we should avoid static functions except for helper functions and singleton pattern instances in ur case.

RajkiranBagal: Okay!

RajkiranBagal (approved):

classes/class-uagb-update.php
```diff
@@ -49,31 +49,53 @@ public function __construct() {
...
+		/**
+		 * Update asset generation option if it is not exist.
+		 *
+		 * @since 1.22.4
+		 * @return void
+		 */
+		public function fresh_install_update_asset_generation_option() {
+
+			if ( UAGB_Helper::is_uag_dir_has_write_permissions() ) {
+				update_option( '_uagb_allow_file_generation', 'enabled' );
```
rohitp-bsf, line 96: One thing we can improve here in the future by wrapping all uagb options inside the `uagb-setting[key]` something like. Which will help to delete and reset in the future.

rohitp-bsf (approved):

classes/class-uagb-update.php
rohitp-bsf: Also, helps to reduce the database calls.

sandeshjangam: Yes, we will consider this.

AbhijitNage123 (approved):

PR #668, opened 2021-05-12
new: Beta & Rollback to Previous Version Support in UAG.

RajkiranBagal:
### Description
 - Trello task - https://trello.com/c/l6OB5mb2/312-improvement-beta-support-in-uag

### Types of changes

### How has this been tested?
 - Added in the task.

### Checklist:
- [x] My code is tested


classes/class-uagb-admin.php
```diff
@@ -527,6 +581,94 @@ public static function toc_plugin( $plugins ) {
...
+		 * Rollback to previous UAG version.
+		 *
+		 * Fired by `admin_post_uag_rollback` action.
+		 *
+		 * @since 1.23.0
+		 * @access public
+		 */
+		public static function post_uagb_rollback() {
+
+			check_admin_referer( 'uag_rollback' );
```
sandeshjangam, line 597: Do we need to check the current user's manage_option here?

RajkiranBagal: No there is no need to check manage_option here as it is a WP hook & is authenticated already - https://developer.wordpress.org/reference/hooks/admin_post_action/

sandeshjangam: @RajkiranBagal It seems that, checked only is_user_logged_in condition
https://core.trac.wordpress.org/browser/tags/5.7.1/src/wp-admin/admin-post.php#L70

RajkiranBagal: Right! Will check the manage_option('adminisatrator') here.

sandeshjangam (approved):

PR #711, opened 2021-05-31
Refactor uag final

rohitp-bsf:
### Description
- Removed extra Lazy loader components as these components were being used in the deprecated files as well.

### Screenshots
- NA

### Checklist:
- [ ] My code is tested


src/blocks/blockquote/render.js
```diff
@@ -1,29 +1,11 @@
 import classnames from 'classnames';
 import UAGB_Block_Icons from '@Controls/block-icons';
-
 import React, { lazy, Suspense } from 'react';
```
RajkiranBagal, line 3: If not needed now then we can also remove these packages.

RajkiranBagal (changes requested):

src/blocks/blockquote/render.js
rohitp-bsf: Yes, that needed to make SCSS lazy. That is intenstional.

RajkiranBagal: Okay!

RajkiranBagal (approved):

PR #756, opened 2021-06-15
Add an action to "broadcast" the common CSS for a single block.

kmgalanakis:
**Please don't merge this, while it's in draft state**

### Description
Adds an action that broadcasts the common CSS for a single block. This is needed in our loops so that we can track the common CSS and append it to the styles only once.


kmgalanakis:
@RajkiranBagal this is ready for review and merging.

vrundakansara (approved):

PR #954, opened 2021-09-06
UAG-385 - Fix: UAG blocks losses style on single post when Auto Load Previous Posts is ON.

AbhijitNage123:
### Description

- https://ua-gutenberg.atlassian.net/browse/UAG-385

### Types of changes

### How has this been tested?

- Create few posts.
- Add the UAG block to the content of those posts.
- Activate "Auto Load Previous Posts" option in Customizer > Blog > Single Post.
- Click on full Post text and scroll down to the next auto-loaded post. You will see that the UAG block doesn't have styling. It doesn't look good. 
- Switch to `uagb-block-losses-style-when-auto-load-pre-post-is-on` Check UAG block asset is loading on the post and all is working fine.
- Disable the `Auto Load Previous Posts` and check again.
- Switch to a different theme and regenerate assets and check post everything looks good.

### Checklist:
- [x] My code is tested


RajkiranBagal (approved):

classes/class-uagb-front-assets.php
```diff
@@ -108,6 +108,23 @@ public function set_initial_variables() {
...
+				'numberposts' => -1,
+			);
+
+			$posts = get_posts( $args ); // Retrieves an array of the posts matching the given criteria.
+
+			foreach ( $posts as $post ) {
+				$current_post_assets = new UAGB_Post_Assets( $post->ID );
+				$current_post_assets->enqueue_scripts();
+			}
+		}
```
sandeshjangam, line 126: Ideally, We should do this on the actions from a separate compatibility file. Move this code into a separate file and only load it when Astra is activated. 

Never add any 3rd party compatibility code directly in your main execution. It should be always on actions and filters.

Create a compatibility folder for all such 3rd part compatibilities.

classes/class-uagb-front-assets.php
```diff
@@ -108,6 +108,23 @@ public function set_initial_variables() {
 	 */
 	public function enqueue_asset_files() {
 
+		$theme_settings_array = ( defined( 'ASTRA_THEME_SETTINGS' ) ) ? get_option( ASTRA_THEME_SETTINGS ) : '';
+
+		/* Compatibility with Astra Auto Load Previous Posts for single post. */
+		if ( isset( $theme_settings_array['ast-auto-prev-post'] ) && is_singular() && 'post' === get_post_type() ) {
+
+			$args = array(
+				'numberposts' => -1,
```
Nikschavan, line 117: Is there any way to avoid `'numberposts' => -1`?

Nikschavan (changes requested):

classes/class-uagb-front-assets.php
sandeshjangam: Yes, we are not going to include this code. As it is not the right way to do this. Trying a more optimized way to load CSS.

AbhijitNage123:
This needs to be fixed by Astra. Added task here - https://wp-astra.atlassian.net/browse/AST-1010

PR #957, opened 2021-09-08
Improvements - Replace `wp.` const to import.

AbhijitNage123:
### Description

- Replace `wp.` const to import.

### Types of changes

- Improvement

### How has this been tested?

- All should be working fine.

### Checklist:
- [x] My code is tested


blocks-config/uagb-controls/block-icons.js
```diff
@@ -2,7 +2,9 @@
  * Block Icons
  */
 
-const el = wp.element.createElement;
+ import {
+	createElement as el
+} from 'react';
```
sandeshjangam, line 7: Can we do something like this?

```suggestion
 import {
	createElement as el
} from '@wordpress/element';
```

RajkiranBagal (approved):

blocks-config/uagb-controls/block-icons.js
AbhijitNage123: Resolved.

sandeshjangam (approved):

PR #1075, opened 2021-10-06
Fix: Call to undefined function `twentyseventeen_panel_count()` when previewing another theme from the customizer.

AbhijitNage123:
### Description

- https://brainstormforce.slack.com/archives/CAQ2TSA6A/p1633516341100000

### Types of changes

- Bug fix ( breaking change which fixes an issue)

### How has this been tested?

- Go to customizer > click on Change > Change theme to Astra or any
- Check Live preview of any theme

### Checklist:
- [x] My code is tested


classes/class-uagb-twenty-seventeen-compatibility.php
```diff
@@ -47,23 +47,25 @@ public function __construct() {
 		 */
 		public function generate_stylesheet() {
 
-			$panel_count     = twentyseventeen_panel_count();
-			$post_assets_obj = uagb_get_front_post_assets();
-			$all_posts       = array();
+			if ( function_exists( 'twentyseventeen_panel_count' ) ) {
```
RajkiranBagal, line 50: Instead of addmin the function in exists  condition, Add the action in condition itself.

RajkiranBagal (changes requested):

classes/class-uagb-twenty-seventeen-compatibility.php
AbhijitNage123: `function_exists( 'twentyseventeen_panel_count' )` Is getting false in action due to which styling break

PR #1175, opened 2021-12-23
Improvement: Added copy-paste styling Attribute value in all the block

supriyajjadhav:
### Description
Set attribute value to true which styling we need to copy in every block

### Types of changes
non-breaking change which adds functionality

### Checklist:
- [x] My code is tested


RajkiranBagal (approved):

akshaylakh:
- Initially, the Paste button should be disabled, the user should not be able to click paste
- Change Tooltip to UAG Copy/Paste
- Are we not going to copy-paste the full block (content)? Full page copy paste. Why only styles? Why not settings?
- Copy-Paste is present in other blocks too? Kadence blocks too.
- As we are supporting copy-paste style, So can we enable the paste option for the same blocks? Ex.ample The user copied the creative heading block so the paste should be enabled for another creative heading block

sandeshjangam (approved):

PR #1191, opened 2022-01-05
Improvement : Units-wise min/max option for Range and Responsive Slider.

AbhijitNage123:
### Description

- Added option to add min/max by unit
- eg. For PX I want max 1600 min 100 and same for % I want max 100 min 10 now it is possible.
- We just need to pass the limit min/max attribute to the Range and Responsive slider.
- eq. `limitMin={ { 'px': 100, '%': 10 } }
					limitMax={ { 'px': 1500, '%': 100 } }`
### Types of changes

- New feature (non-breaking change which adds functionality)

### How has this been tested?

- For Responsive Slider - Drag and drop TOC
- Go to the General > Content Tab > Check Gap Between Lists setting for PX max is 1600px and for % max is 100%.
- For Range - Drag and drop Info-box
- Go to Style Tab > Icon/Image tab > Size setting for PX max is 1600px and for EM max is 100em

### Checklist:
- [x] My code is tested


src/components/range/Range.js
```diff
@@ -141,6 +142,9 @@ const Range = ( props ) => {
 		return items;
 	};
 
+	const max = ( props.limitMax ) ? limitMax( props.unit?.value, props ) : props.max;
```
RajkiranBagal: Can we write just `const max = limitMax( props.unit?.value, props )` here? as we are checking necessary conditions in the function. Same applies everywhere.

RajkiranBagal (changes requested):

src/components/range/Range.js
AbhijitNage123: Updated.

src/components/responsive-slider/index.js
```diff
@@ -1,28 +1,19 @@
...
-		undefined !== props.data.mobile.max ? props.data.mobile.max : props.max;
-	const minDesk =
-		undefined !== props.data.desktop.min
-			? props.data.desktop.min
-			: props.min;
-	const minTab =
-		undefined !== props.data.tablet.min ? props.data.tablet.min : props.min;
-	const minMob =
-		undefined !== props.data.mobile.min ? props.data.mobile.min : props.min;
+	const isResponsive = true;
```
RajkiranBagal: Instead of declaring a variable, directly pass true

RajkiranBagal (changes requested):

src/components/responsive-slider/index.js
AbhijitNage123: Done

RajkiranBagal (approved):

akshaylakh:
Working fine

sandeshjangam (approved):

PR #1250, opened 2022-02-07
Improvement: Added Variant support for Typography control.

supriyajjadhav:
### Description
- Load variant as per user selection to improve typography performance. refer this screencast - 
https://share.getcloudapp.com/P8u8RBmL
Apply only when you select
Preview in editor
For eq.
If we have font-weight like 100,200 and variant italic. So it will need to load variants like 100,100italic, 200, 200italic, etc. And URL decode look like this -
 http://fonts.googleapis.com/css?family=Crafty
 Girls:400,400italic|Sorts Mill Goudy:400,400italic&subset=latin&display=fallback&ver=2.0.0-beta.1 

### Types of changes
breaking change which adds functionality

### Checklist:
- [ ] My code is tested


classes/class-uagb-helper.php
```diff
@@ -827,6 +843,7 @@ public static function get_typography_css( $attr, $slug, $selector, $combined_se
 			$transform_slug  = ( '' === $slug ) ? 'fontTransform' : $slug . 'Transform';
 			$decoration_slug = ( '' === $slug ) ? 'fontDecoration' : $slug . 'Decoration';
 			$style_slug      = ( '' === $slug ) ? 'fontStyle' : $slug . 'FontStyle';
+			$FontVariant     = ( '' === $slug ) ? 'fontVariant' : $slug . 'fontVariant';
```
RajkiranBagal, line 846: PHP variables should always be in snake_case

classes/class-uagb-helper.php
```diff
@@ -225,14 +225,30 @@ public static function get_css_value( $value = '', $unit = '' ) {
 		 * @param array $load_google_font the blocks attr.
 		 * @param array $font_family the blocks attr.
 		 * @param array $font_weight the blocks attr.
+		 * @param array $font_variant the blocks attr.
 		 */
-		public static function blocks_google_font( $load_google_font, $font_family, $font_weight ) {
+		public static function blocks_google_font( $load_google_font, $font_family, $font_weight, $font_variant = '' ) {
+
+			if( isset( $font_variant ) ){
```
RajkiranBagal, line 232: If $font_variant is empty then do not do any below Font Variant processing.

src/components/typography/font-typography.js
```diff
@@ -123,17 +159,51 @@ function FontFamilyControl( props ) {
 			height: '30px',
 		} ),
 	}
-	
+	const customSelectVariant = {
+		container: ( provided ) => ( {
```
RajkiranBagal, line 163: Where is this exactly used?

RajkiranBagal (changes requested):

classes/class-uagb-helper.php
supriyajjadhav: if font_variant is empty then we store font-weight in the same array. so we can pass this array to generate fonts URL

src/components/typography/font-typography.js
supriyajjadhav: Removed.

RajkiranBagal (approved):

PR #1354, opened 2022-03-07
fix: WP Query gets called on every refresh.

RajkiranBagal:
### Description
 - WP Query gets called on every refresh for displaying notices.

### Types of changes
 Bugfix (non-breaking change which fixes an issue) 

### How has this been tested?
 - The notice function should be called only on Post Save/Update.

### Checklist:
- [x] My code is tested


classes/class-uagb-rest-api.php
```diff
@@ -93,6 +93,47 @@ public function delete_page_assets( $post_id ) {
...
+		}
+
+		/**
+		 * Get flag if more than 5 pages are build using UAG.
+		 *
+		 * @since x.x.x
+		 */
+		public function check_for_adding_notice() {
+
+			if ( false === $posts_created_with_uag || 5 > $posts_created_with_uag ) {
```
sandeshjangam, line 107: $posts_created_with_uag  undefined variable

RajkiranBagal: Right! After testing It got removed along with the var_dump. Added it.

classes/class-uagb-rest-api.php
```diff
@@ -93,6 +93,49 @@ public function delete_page_assets( $post_id ) {
...
+					foreach ( $query->posts as $key => $post ) {
+						if ( $uag_post_count >= 5 ) {
+							break;
+						}
+
+						if ( false !== strpos( $post->post_content, '<!-- wp:uagb/' ) ) {
+							$uag_post_count++;
+						}
+					}
+				}
```
sandeshjangam, line 131: This will create the same problem for saving which user is facing in the admin. Check if the current post is created with UAG then update the count +1. If it is already greater than our threshold breaks it.

sandeshjangam (changes requested):
Need to update logic

classes/class-uagb-rest-api.php
RajkiranBagal: Updated, Can you please check now

PR #1506, opened 2022-05-02
improvement: Optimized Icon loading by dividing large icons file into chunks & then loading them.

RajkiranBagal:
### Description
https://app.startinfinity.com/b/Mi5zu3VRVzk/M7Dpq9yR4Pe/d8882757-e1aa-46dc-8d91-ccdf2b30b4fa?redirect=true&view=0ce18087-e6b2-4a91-b54c-c386ce1f811c&t=attributes

### Types of changes

### How has this been tested?
 - In the card.

### Checklist:
- [x] My code is tested


sandeshjangam (approved):

classes/class-uagb-helper.php
```diff
@@ -282,6 +290,7 @@ public static function render_svg_html( $icon ) {
 			$icon = sanitize_text_field( esc_attr( $icon ) );
 
 			$json = self::backend_load_font_awesome_icons();
+			$json = array_merge( $json[0], $json[1], $json[2] );
```
sandeshjangam, line 293: Need to confirm performance here

sandeshjangam (changes requested):
Need to confirm performance

blocks-config/uagb-controls/getUAGBIcons.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Get Icons Array.
+ *
+ * @return {Array} Get Icons Array.
+ */
+
+function getUAGBIcons() {
+	return { ...uagb_blocks_info.uagb_svg_icons_1, ...uagb_blocks_info.uagb_svg_icons_2, ...uagb_blocks_info.uagb_svg_icons_3 }
+}
```
Nikschavan, line 9: Since we are localizing and then loading the icons in JS, Can we directly require the JSON file in JS?
something like - 

```JS
import icons from '../assets/icons.json';
```

because right now I think we are processing and loading all icons in PHP and then anyways loading in JS.
This approach will bundle the icons file in build JS though, so build file size will increase a lot. Right now though all these icons are loaded in JS anyways, but ot from the build file, but from localized which adds that extra size increase in the dom instead of the JS file.

Do let me know your thoughts @sandeshjangam / @RajkiranBagal

PR #1552, opened 2022-05-17
fix: Team Block icon style

jamyrbsf:
### Description
When we add Team, if we increase the size of social icons it does but if we want to reset the size as default then the option for reset is not working

Icon spacing was not working for mobile and tablet in the editor.

### Types of changes
Icon Style

### How has this been tested?
By coded

### Checklist:
- [ ] My code is tested


src/blocks/team/settings.js
```diff
@@ -1052,35 +1051,29 @@ const Settings = ( props ) => {
...
-						value: socialFontSizeMobile,
-						label: 'socialFontSizeMobile',
-					} }
-					fontSizeTablet={ {
-						value: socialFontSizeTablet,
-						label: 'socialFontSizeTablet',
-					} }
-					sizeText={ __(
+				<ResponsiveSlider
+					label={__(
```
RajkiranBagal, line 1055: @jamyrbsf You missed the socialFontSizeType, What about that? Can you check?

RajkiranBagal (changes requested):

src/blocks/team/settings.js
jamyrbsf: We do not need it in Responsive Slider Previously it was used in TypographyControl.

jamyrbsf: and it is set to default px

PR #1596, opened 2022-05-25
SPCTR-862 Button Block: Option to Add BG Color for Hover When Transparent/Gradient BG Selected

sushmak02:
### Description
In the Button block, when we select Transparent or Gradient as the background option, we do not have an option to set background colors for the hover state. IMO, we must have hover options but the Button to bring the design flexibility.

### Screenshots
Refer - https://share.getcloudapp.com/9ZuoGEPk

### Types of changes

### How has this been tested?
Test all the hover and normal color combinations

### Checklist:
- [x] My code is tested


src/blocks/buttons-child/attributes.js
```diff
@@ -154,11 +154,18 @@ const attributes = {
 			styleType: 'btn-padding-unit'
 		}
 	},
-	gradientValue: {
```
RajkiranBagal, line 157: We cannot change previous attributes names, because of backward compatibility issues. But Yes we cal always add new attributes if needed.

RajkiranBagal (changes requested):

RajkiranBagal (approved):

PR #2304, opened 2022-09-14
SPCTR-1801 Fix: Responsive condition option does not save when hide on multiple devices is selected.

AbhijitNage123:
### Description

- https://brainstormforce.atlassian.net/browse/SPCTR-1801

### Screenshots

- https://share.getcloudapp.com/GGuzojwg

### Types of changes

### How has this been tested?

- Added to the card.

### Checklist:
- [x] My code is tested


src/blocks/extensions/advanced-settings/index.js
```diff
@@ -187,7 +187,6 @@ const ResponsiveConditionOptions = ( props ) => {
 				onChange={ () =>
 					setAttributes( {
 						UAGHideDesktop: ! attributes.UAGHideDesktop,
-						UAGResponsiveConditions: ! UAGResponsiveConditions
```
RajkiranBagal, line 190: This is needed for old users.

RajkiranBagal (changes requested):

src/blocks/extensions/advanced-settings/index.js
AbhijitNage123: It is causing settings are not saved

AbhijitNage123: Updated logic. Please verify

AbhijitNage123: @RajkiranBagal As discussed, I have tested the case and it does not adding any class to block whenever toggle is set to off. Cause we have added condition inside of if again to check is if toggle is true then only add the class to the block. Screencast - https://share.getcloudapp.com/v1uQ2b79

RajkiranBagal (approved):

PR #2322, opened 2022-09-21
SPCTR-1851 - Fix: FAQ - Throwing an undefined style issue in console when html anchor is used.

supriyajjadhav:
### Description
Fix: FAQ - Throwing an undefined style issue in console when html anchor is used.

### Screenshots
Testing Evidence - https://share.getcloudapp.com/z8uQPYD6

### Types of changes

### Checklist:
- [x] My code is tested


assets/js/faq.js
```diff
@@ -80,10 +81,10 @@ function setupFAQ() {
 				.classList.add( 'uagb-faq-item-active' );
 			elementToOpen
 				.setAttribute( 'aria-expanded', true );
-			slideDown(
-				elementToOpen.getElementsByClassName( 'uagb-faq-content' )[ 0 ],
-				500
-			);
+			const faqContent =  elementToOpen.getElementsByClassName( 'uagb-faq-content' )[0];
```
AbhijitNage123, line 84: Can we use directly below code instead of adding extra condition.
`elementToOpen?.getElementsByClassName( 'uagb-faq-content' )[ 0 ],
				500
			);`

AbhijitNage123 (changes requested):

assets/js/faq.js
supriyajjadhav: @AbhijitNage123 This will return undefined in the function parameter, So we need to check first if this is not undefined, then only we can call the function.

AbhijitNage123 (commented):
@supriyajjadhav Need the same condition for all slideDown as well as slideUp functions. Please update it accordingly. Seem like you only update at one line

AbhijitNage123 (approved):

supriyajjadhav:
> @supriyajjadhav Need the same condition for all slideDown as well as slideUp functions. Please update it accordingly. Seem like you only update at one line

@AbhijitNage123 Tested all the possible cases. We no need to add conditions everywhere.

AbhijitNage123 (approved):

assets/js/faq.js
```diff
@@ -80,10 +81,10 @@ function setupFAQ() {
...
+			if( faqContent !== undefined ){
```
RajkiranBagal, line 85: Can we write `if ( faqContent ) {`

RajkiranBagal (changes requested):

assets/js/faq.js
supriyajjadhav: yes, updated code with suggested changes

RajkiranBagal (approved):

RajkiranBagal (approved):

RajkiranBagal (approved):

PR #2481, opened 2022-11-21
v2.0.16

RajkiranBagal:
* Improvement: Spectra Core Blocks - Keep core blocks always enabled & prevent them for being disabled.
* Fix: The "uagb_spectra_font_awesome_polyfiller" AJAX gets called multiple times.
* Fix: Post Blocks: The "spectra/v1/all_taxonomy" Rest API gets called multiple times inside editor.
* Fix: Editor layout shifts to right when clicking on any block settings.
* Fix: Spectra component styling gets overridden in the customizer block editor.
* Fix: Info Box - The title is disabled on the editor if we set P tag.
* Fix: Team - Social icon style option is showing when the social icon is disabled.
* Fix: Tabs - Layout fixed when preset is selected.
* Fix: Buttons - Validation errors when we add `Add "nofollow" to link.


imrantushar (approved):

jamyrbsf (approved):

aapatre (approved):

admin-core/assets/src/store/globalDataStore.js
```diff
@@ -32,6 +32,14 @@ const initialState = {
     enableFontAwesome5: ( 'yes' === uag_react.global_data.uagb_old_user_less_than_2 ) ? 'enabled' : 'disabled',
     enableAutoBlockRecovery: ( 'yes' === uag_react.global_data.uagb_old_user_less_than_2 ) ? 'enabled' : 'disabled',
 	enableLegacyBlocks: ( 'yes' === uag_react.global_data.uagb_old_user_less_than_2 ) ? 'yes' : 'no',
+    coreBlocks: [
```
patilvikasj, line 35: This array is defined at two different places. Can we declare it at one place and fetch it at another place? Like maybe using localise_script.

raojip (approved):

supriyajjadhav (approved):

bsf-zanev (approved):
Good to go.

admin-core/assets/src/store/globalDataStore.js
RajkiranBagal: It is just added as an initial loading array. After the Redux store is initialized the array in the PHP file gets assigned to the key. Adding localized variables will add more processes to it. So for now just added a static array.

Even if we remove that everything will work fine, but added just to be sure.

patilvikasj (approved):

PR #2500, opened 2022-11-24
SPCTR-2043 - freemium features

imrantushar:
### Checklist:
- [ ] My code is tested


src/components/icon-picker/index.js
```diff
@@ -1,10 +1,20 @@
...
 import { __ } from '@wordpress/i18n';
 
 const UAGIconPicker = ( props ) => {
+	const [panelNameForHook, setPanelNameForHook] = useState( null );
+	const panelRef = useRef( null );
+
+	const { getSelectedBlock } = select( 'core/block-editor' );
+	useEffect( () => {
+		setPanelNameForHook( getPanelIdFromRef( panelRef ) )
+	}, [blockNameForHook] )
```
abhishekrijal, line 16: This will cause an issue as `blockNameForHook` is defined in L:27 and used here.

src/components/background/index.js
```diff
@@ -1,9 +1,10 @@
 import { __ } from '@wordpress/i18n';
+import { useRef } from '@wordpress/element';
 import AdvancedPopColorControl from '@Components/color-control/advanced-pop-color-control.js';
 import { SelectControl } from '@wordpress/components';
 import styles from './editor.lazy.scss';
 import GradientSettings from '@Components/gradient-settings';
-import React, { useLayoutEffect } from 'react';
+import React, { useLayoutEffect, useEffect, useState } from 'react';
```
abhishekrijal, line 7: Not sure why it is used in this way, but wouldn't it be better if we use these types of hooks from `@wordpress/element` directly ?

includes/blocks/advanced-heading/frontend.css.php
```diff
@@ -268,9 +268,9 @@
 );
 
 $combined_selectors = array(
-	'desktop' => $selectors,
-	'tablet'  => $t_selectors,
-	'mobile'  => $m_selectors,
+	'desktop' => apply_filters( 'spectra_' . $block_name . '_styling', $selectors, $attr ),
+	'tablet'  => apply_filters( 'spectra_' . $block_name . '_tablet_styling', $t_selectors, $attr ),
+	'mobile'  => apply_filters( 'spectra_' . $block_name . '_mobile_styling', $m_selectors, $attr ),
```
abhishekrijal, line 273: This type of filter is required for frontend.js.php also. Also, Can we generalize this filter in some way so that we don't have to write it in every block's `frontend.css.php`?

src/components/reset/index.js
```diff
@@ -77,23 +85,33 @@ const UAGReset = ( props ) => {
 		}
 	};
 
+	const controlName = 'reset'; // there is no label props that's why keep hard coded label
+	const controlBeforeDomElement = wp.hooks.applyFilters( `spectra.${blockNameForHook}.${panelNameForHook}.${controlName}.before`, '', blockNameForHook );
```
adi3890, line 89: Can we use component to avoid code redundancy in all blocks

includes/blocks/advanced-heading/frontend.css.php
adi3890: Having helper function would be better approach

src/components/icon-picker/index.js
abhishekrijal: Fixed

includes/blocks/advanced-heading/frontend.css.php
abhishekrijal: With the current structure, filters can only be added to the file itself. Leaving it to remove blockers for now.

RajkiranBagal (approved):

includes/blocks/advanced-heading/frontend.css.php
```diff
@@ -268,9 +268,9 @@
 );
 
 $combined_selectors = array(
-	'desktop' => $selectors,
-	'tablet'  => $t_selectors,
-	'mobile'  => $m_selectors,
+	'desktop' => apply_filters( 'spectra_' . $block_name . '_styling', $selectors, $attr ),
```
aapatre, line 271: Since we are following this pattern for all blocks, I've implemented the same in countdown.

aapatre (approved):

adi3890 (approved):

patilvikasj (approved):

PR #2623, opened 2023-01-12
Spectra V2.3.0

RajkiranBagal:
* New: Introducing the Slider block that allows you to make almost anything into a carousel, a good way to grab attention!
* Improvement: Added support for Astra's 4.0 theme menu position next to the Dashboard in admin.
* Fix: Container - Adding padding to paragraph cause uneven space and cross the boundary of container on responsive devices.
* Fix: Container - Background video goes out the boundary of the container if we have rounded corner.
* Fix: Image Gallery - When adding any of the HTML code or adding the link, the visibility of that caption gets changed to always visible
* Fix: Image Gallery - Broken links were shown in the gallery on frontend when added to caption.
* Fix: Icon List - Disable link for the list item, it turns on the toggle again.
* Fix: Testimonial - Carousel dots overlapping on below blocks in the mobile device.
* Fix: Post Grid - Undefined array key issue.
* Fix: How-To - Image is not visible when image selected from Image control.
* Fix: Image - Does not wrap with paragraph.
* Fix: Tabs - Container get hidden inside tabs body whenever use.
* Fix: Global - Buttons - Removed the WP default allowedFormats from the toolbar.
* Fix: Global - SVG Icons for blocks look big after switching to responsive devices inside the editor.
* Deprecated: Global - Removed deprecated toolbar and multiline for rich text.


includes/blocks/slider-child/attributes.php
```diff
@@ -0,0 +1,86 @@
+<?php
+/**
+ * Attributes File.
+ *
+ * @since 2.0.0
```
AbhijitNage123, line 5: it should be 2.3.0

includes/blocks/slider-child/block.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * Block Information.
+ *
+ * @since 2.0.0
```
AbhijitNage123, line 5: Same here should be 2.3.0

includes/blocks/slider-child/frontend.css.php
```diff
@@ -0,0 +1,154 @@
+<?php
+/**
+ * Slider child front end style
+ *
+ * @since 2.0.0
```
AbhijitNage123, line 5: same here

includes/blocks/slider/attributes.php
```diff
@@ -0,0 +1,135 @@
+<?php
+/**
+ * Attributes File.
+ *
+ * @since 2.0.0
```
AbhijitNage123, line 5: same here

includes/blocks/slider/frontend.js.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * Frontend JS File.
+ *
+ * @since 2.0.0
```
AbhijitNage123, line 5: same here

patilvikasj (approved):

src/blocks/slider-child/save.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * BLOCK: Container - Save Block
```
AbhijitNage123, line 2: change name to slider

AbhijitNage123 (approved):

PR #2730, opened 2023-02-13
Refactor: Rewritting get_css_value php function

adi3890:
### Description
**Rewriting code for get_css_value**.
Previously this function behaviors for PHP 8.1+ and lower versions as we where user  == ( loose comparison )
Current logic checks if value is numeric then return value suffixed with unit
or else returns '' ( empy string ) 

### Checklist:
- [ ] My code is tested


classes/class-uagb-helper.php
```diff
@@ -202,18 +202,11 @@ public static function generate_css( $selectors, $id ) {
...
-
-			if ( '' == $value ) { // phpcs:ignore WordPress.PHP.StrictComparisons.LooseComparison
-				return $value;
-			}
-
-			$css_val = '';
-
-			if ( isset( $value ) ) {
-				$css_val = esc_attr( $value ) . $unit;
+			if ( ! is_numeric( $value ) ) {
```
patilvikasj, line 205: Can this value be text also? In case of typography options?

AbhijitNage123 (approved):

classes/class-uagb-helper.php
adi3890: No this works for any numeric value only.
like 0, 0.25, '35' etc.

I am not sure but seems like for typography we have function get_typography_css

RajkiranBagal: We use get_css_value function for applying units to numeric values like padding, margin, width, font size , line-height, so I think it's okay.

patilvikasj (approved):

classes/class-uagb-helper.php
adi3890: @patilvikasj is there any other use case where we use this function

RajkiranBagal (approved):

abhishekrijal (approved):
Looks good to me

classes/class-uagb-helper.php
```diff
@@ -202,18 +202,11 @@ public static function generate_css( $selectors, $id ) {
...
+				return '';
 			}
 
-			return $css_val;
+			return esc_attr( $value ) . $unit;
```
bsf-zanev, line 209: Should we add the case of rendering 0 without units? ( As followed in CSS conventions )

Pros: Rendering 0 for outline, border, etc is better than rendering 0px or 0em for the same.
Cons: 0 can't be used in the calc() function without the units - then again, 0 shouldn't be in a calc function in most cases.

I feel we should exclude the units for 0, and be vary when using calc(). Let me know your thoughts.

@patilvikasj, @RajkiranBagal, @adi3890.

patilvikasj: Maybe we can handle that while calling `get_css_value` function. Pass the unit as blank when it is not going to be used in calc. @bsf-zanev

RajkiranBagal: @bsf-zanev Okay! The case you are saying is a static case that contains a static 0 for CSS with no units, in that case this function should not be used & 0 should be added directly in the CSS.

bsf-zanev: Exactly! But this affects dynamic as well - if the user is using a slider that can go to 0, or a spacing control with a 0 value, the Zeros should not use a unit in the css.

adi3890: it can create an issue with properities like transition related css.

bsf-zanev: True, and besides that we can avoid the ternary rendering of units for calc() if we keep the code as is.

bsf-zanev (approved):
Looks good to me.

supriyajjadhav (approved):

RajkiranBagal (approved):

PR #2743, opened 2023-02-19
SPCTR-2378-code-rafactor-useeffect-dependency

rahulvarma722:
1. Refactor useEffect dependency in all blocks
2. Now we should pass only attributes instead of whole props for styling.


aapatre (approved):

aapatre:
Approved with changes:
- Please update the description to explain the changes in details
- Like: Added isSelected to display settings only when block is selected and optimize the same

aapatre (approved):
LGTM!

RajkiranBagal (approved):

PR #2775, opened 2023-03-07
SPCTR-2437 - Implementing prettie formatter

rahulvarma722:
### Description
Prettier has been configured in the plugin so the inconsistency of code in JS files and CSS files will be now resolved.


.prettierrc.js
```diff
@@ -1,3 +1,15 @@
...
 // Useful for editor integrations.
-module.exports = require( '@wordpress/prettier-config' );
\ No newline at end of file
+const config = require( '@wordpress/prettier-config' );
+
+config.overrides = [
+	{
+		files: [ '*.scss', '*.css' ],
+		options: {
+			printWidth: 500,
```
sandeshjangam, line 9: We can increase this?

assets/css/blocks/buttons.css
```diff
@@ -1 +1,47 @@
...
+.uagb-buttons__outer-wrap .uagb-buttons__wrap {
```
adi3890, line 1: Please check if these files should be excluded in linting

rahulvarma722: excluded

.prettierrc.js
rahulvarma722: 800 done

PR #2787, opened 2023-03-10
SPCTR-1911 Countdown Pro: countdown-pro-next-release to next-release PR

aapatre:
### Description
- This PR contains evergreen timer and expiry actions related code for Spectra Free/Lite plugin

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code is tested


aapatre:
Looks good to me:
- Checked backward compatibility by creating countdowns in next-release and then switching to this branch
- Also on multiple instances, I checked the behavior of the block when some pro features are used (evergreen, expiry actions) and Spectra Pro is disabled: no breaks as such, especially on frontend

The only edge case break in editor would be when the user chooses 'Replace' expiry action (which contains innerblocks) and then disable Spectra Pro, they would expect 'Attempt Block Recovery' issue (which is intended since legally they are allowed to use innerblocks only when Pro is activated.

For all other expiry actions, when Spectra Pro is disabled, the countdown remains at zero after it expires.

RajkiranBagal (approved):

src/blocks/countdown/edit.js
```diff
@@ -110,8 +113,14 @@ const UAGBCountdownEdit = ( props ) => {
 		deviceType
 	] );
 
+	const countdownProToolbar = wp.hooks.applyFilters( 'spectra.countdown.toolbar-hook', '', props.name );	
+
 	return (
 		<>
+			{/* Countdown Toolbar options for Pro */}
+			{ ( !!uagb_blocks_info.spectra_pro_status && props.attributes.timerEndAction === 'content' ) &&
+				countdownProToolbar
```
adi3890, line 122: can we directly apply filter here.
importing wordpress/hooks will be good

src/blocks/countdown/render.js
```diff
@@ -36,15 +37,28 @@ const Render = ( props ) => {
 		ref: countdownRef
 	} );
 
+	const innerblocks_content = wp.hooks.applyFilters( 'spectra.countdown.render-innerblocks', '', props.name );
+
+	const innerblocks_structure = ( ( !!uagb_blocks_info.spectra_pro_status && timerEndAction === 'content' ) &&
+		<div
+			className={ `uagb-block-countdown-innerblocks-${ block_id } wp-block-uagb-countdown-innerblocks` }
+		>
+			{innerblocks_content}
```
adi3890, line 46: do we need to create this constant cant we apply filter here directly

src/blocks/countdown/save.js
```diff
@@ -11,22 +11,36 @@ export default function Save( props ) {
...
 		className: `uagb-block-${ block_id } wp-block-uagb-countdown`
 	} );
 
+	const innerblocks_content = wp.hooks.applyFilters( 'spectra.countdown.save-innerblocks', '', props.name );
+
+	const innerblocks_structure = ( ( !!uagb_blocks_info.spectra_pro_status && timerEndAction === 'content' ) &&
+		<div
+			className={ `uagb-block-countdown-innerblocks-${ block_id } wp-block-uagb-countdown-innerblocks` }
+		>
+			{innerblocks_content}
```
adi3890, line 29: same as above

assets/js/uagb-countdown.js
```diff
@@ -1,6 +1,9 @@
+// Global uagb_countdown_data.
 UAGBCountdown = { // eslint-disable-line no-undef
```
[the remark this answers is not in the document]
aapatre: We need this in global scope, and if we add `let` or `const` the scope won't be global.

aapatre:
@adi3890 I've resolved the above comments

adi3890 (approved):

rahulvarma722 (approved):

RajkiranBagal (approved):

PR #2847, opened 2023-03-22
Added filter for disable responsive condition for theme

AbhijitNage123:
### Description

- Added filter to disable/enable display condition for core.

### Screenshots

- NA

### Types of changes

- NA

### How has this been tested?

- NA

### Checklist:
- [x] My code is tested


src/blocks/extensions/condition-block/index.js
```diff
@@ -223,8 +224,7 @@ const AdvancedControlsBlock = createHigherOrderComponent( ( BlockEdit ) => {
 						>
 							{ __( 'Filter to disable responsive/display condition. ' ) }
 						</ExternalLink>
-						<hr className="uagb-editor__separator" />
-						{ UserResponsiveConditionOptions( props ) }
+						{ '1' === enableResponsiveConditionsForCoreBlocks && UserResponsiveConditionOptions( props ) }
```
RajkiranBagal, line 227: Why we have checked '1'? '1' === true gives false.

RajkiranBagal (changes requested):

src/blocks/extensions/condition-block/index.js
AbhijitNage123: By default `enableResponsiveConditionsForCoreBlocks` is true which return in js as 1 so we are showing responsive condition only when it is 1 else hide it. @RajkiranBagal

RajkiranBagal: Have you confirmed this case? @AbhijitNage123

AbhijitNage123: @RajkiranBagal Yes confirm when it set to false it return empty and for true it set to 1

PR #2924, opened 2023-04-03
Fix: PHP Explode Parameter Two Handlers

bsf-zanev:
### Description
Updated PHP Explode Parameter Two Checks to ensure no null value passed in place of string.

### Types of changes

### How has this been tested?
Tested after regenerating assets in the front-end, as well as with matrix alignment for Image Gallery.

### Checklist:
- [x] My code is tested
- [x] My code follows accessibility standards 

### DIFF
--- a/classes/class-uagb-block-helper.php
+++ b/classes/class-uagb-block-helper.php
@@ -1732,7 +1732,7 @@ if ( ! class_exists( 'UAGB_Block_Helper' ) ) {
 		public static function get_matrix_alignment( $value, $pos, $format = '' ) {
 
 			// Return early if remote styles is not a string, or is empty.
-			if ( ! is_string( $value ) || empty( $value ) ) {
+			if ( ! is_string( $value ) || empty( $value ) || ! is_int( $pos ) ) {
 				return '';
 			}
 
--- a/classes/class-uagb-block-helper.php
+++ b/classes/class-uagb-block-helper.php
@@ -1731,8 +1731,8 @@ if ( ! class_exists( 'UAGB_Block_Helper' ) ) {
 		 */
 		public static function get_matrix_alignment( $value, $pos, $format = '' ) {
 
-			// Return early if remote styles is not a string.
-			if ( ! is_string( $value ) ) {
+			// Return early if remote styles is not a string, or is empty.
+			if ( ! is_string( $value ) || empty( $value ) ) {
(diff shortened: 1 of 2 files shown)


RajkiranBagal (approved):

adi3890 (approved):

lib/uagb-webfont/uagb-webfont-loader.php
```diff
@@ -440,13 +440,23 @@ public function get_local_files_from_css() {
 		 */
 		public function get_remote_files_from_css() {
 
+			// Return early if remote styles is not a string, or is empty.
+			if ( ! is_string( $this->remote_styles ) || empty( $this->remote_styles ) ) {
+				return array();
+			}
+
 			$font_faces = explode( '@font-face', $this->remote_styles );
```
adi3890, line 448: should we check for array too.
explode may return false or error for PHP 8.0

bsf-zanev: Yes, I've updated the same here and in the Matrix Alignment as well.

adi3890: lets check for empty value and other **explode** used in this function

classes/class-uagb-block-helper.php
```diff
@@ -1725,10 +1725,26 @@ public static function get_fallback_number( $current_value, $key, $block_name )
...
+
+			// Return early if remote styles is not a string, or is empty.
+			if ( ! is_string( $value ) || empty( $value ) ) {
+				return '';
+			}
+
+			$alignment_array = explode( ' ', esc_attr( $value ) );
+
+			// Return early if alignment propery at the given position is not a string, or is empty.
+			if ( ! is_string( $alignment_array[ $pos - 1 ] ) || empty( $alignment_array[ $pos - 1 ] ) ) {
```
adi3890, line 1742: we should have check for $pos as we are checking for $value

lib/uagb-webfont/uagb-webfont-loader.php
bsf-zanev: This has been done for all the explodes that needed to be updated in Spectra.

adi3890 (approved):

RajkiranBagal (approved):

PR #2928, opened 2023-04-05
SPCTR-2572 Countdown - Countdown not working in Slider block when the Slider is set to infinite loop (on Frontend)

aapatre:
### Description
Countdown not working in Slider block when the Slider is set to infinite loop:
- [JIRA](https://brainstormforce.atlassian.net/browse/SPCTR-2572)
- Drop a slider block and add countdown on the first slide
- Set slider to infinite loop
- On frontend, you can initially see the timer working fine
- But switch slides further, until you come across the first one again
- This time, the countdown is not initialized

This is because the Swiper slider library duplicates the first slide in case it's set to have an infinite loop. Hence the second countdown remains uninitialized.

### Types of changes

### Checklist:
- [x] My code is tested


patilvikasj (approved):

abhishekrijal (approved):
Looks good to me, tested it with slider and independent countdown use case also

RajkiranBagal (approved):

PR #2990, opened 2023-04-15
added css iframe logic

rahulvarma722:
### Description
Added css static logic in iframe
### Checklist:
- [ ] My code is tested


AbhijitNage123 (approved):

blocks-config/uagb-controls/AddStaticStyles.js
```diff
@@ -0,0 +1,14 @@
...
+	const WrapWithStyle = ( props ) => {
+		useEffect( () => {
+			addBlockEditorDynamicStyles();
+		}, [] );
+
+		return <ChildComponent { ...props }/>
+	}
+
+    return WrapWithStyle;
+}
```
supriyajjadhav, line 13: can we return like this?
const MemoizedComponent = React.memo(WrapWithStyle);
return MemoizedComponent;

blocks-config/uagb-controls/AddStaticStyles.js
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from '@wordpress/element';
+import addBlockEditorDynamicStyles from '@Controls/addBlockEditorDynamicStyles';
+const AddStaticStyles = ( ChildComponent )=> {
+	const WrapWithStyle = ( props ) => {
+		useEffect( () => {
+			addBlockEditorDynamicStyles();
+		}, [] );
+
+		return <ChildComponent { ...props }/>
+	}
```
supriyajjadhav, line 10: can we pass ChildComponent like this in param?

}, [ChildComponent]); 

If ChildComponent changes then only the `WrapWithStyle` function will render.

blocks-config/uagb-controls/AddStaticStyles.js
rahulvarma722: This is a higher-order component so we do not need to memoize this component and we are not adding a memo on the Edit component we are adding only a memo in the sub-component.

blocks-config/uagb-controls/AddStaticStyles.js
rahulvarma722: In useEffect,  we add a variable as a dependency but ChildComponent is a component.

blocks-config/uagb-controls/AddStaticStyles.js
supriyajjadhav: Okay. Thank you for the clarification.

RajkiranBagal (approved):

blocks-config/uagb-controls/AddStaticStyles.js
supriyajjadhav: Okay

supriyajjadhav (approved):

PR #3026, opened 2023-04-24
Fix: Heading does not take full width if it's added in the Container which is the inner block of the Countdown block in the Editor.

RajkiranBagal:
### Description
Fix: Heading does not take full width if it's added in the Container which is the inner block of the Countdown block in the Editor.

### Types of changes

### How has this been tested?
 - Add Heading in Container inside Countdown, It should take full width in the editor.

### Checklist:
- [x] My code is tested


mohitsbsftester:
Good to go for container.

PR #3050, opened 2023-04-27
SPCTR- 2764 Loop Builder - Added filter to allowed blocks inside container when used inside loop builder

abhishekrijal:
### Description
Loop Builder - Added filter to allowed blocks inside container when used inside loop builder

### Types of changes

Bug fix (non-breaking change which fixes an issue) -->

### Checklist:
- [ ] My code is tested


RajkiranBagal (approved):

src/blocks/container/render.js
```diff
@@ -110,6 +110,24 @@ const Render = ( props ) => {
 		innerBlocksParams.allowedBlocks = ALLOWED_BLOCKS;
 	}
 
+	const { getBlockParentsAll, getBlockSingle } = useSelect( ( selectStore ) => {
+		const { getBlockParents, getBlock } = selectStore( 'core/block-editor' );
```
RajkiranBagal, line 114: Can we combine above slider loop & this loop? @abhishekrijal @adi3890

RajkiranBagal (changes requested):

src/blocks/container/render.js
abhishekrijal: Slider loop is using the attribute `hasSliderParent ` to get the parent value, in case of loop builder that is achieved with hook directly, while it is better approach, I'll need suggestion from @patilvikasj to modify it, as it might be used somewhere else as well.

RajkiranBagal (approved):

PR #3087, opened 2023-05-04
SPCTR-2809 Fix for container and image broken and wrong placeholder image

abhishekrijal:
### Description
Fix for container and image broken and wrong placeholder image

### Types of changes

### How has this been tested?
Tested by adding the empty image in case of default image and removing it. In frontend the image tag should not be visible when no image URL is set.

### Checklist:
- [ ] My code is tested


src/blocks/image/save.js
```diff
@@ -32,19 +32,21 @@ export default function Save( props ) {
...
-			sizes="(max-width: 480px) 150px"
-			src={ url }
-			alt={ alt }
-			className={ id ? `uag-image-${ id }` : null }
-			width={ width ? width : naturalWidth }
-			height={ height ? height : naturalHeight }
-			title={ title }
-			loading="lazy"
-		/>
+		url && '' !== url && (
```
RajkiranBagal, line 35: I think we do not need to check `'' !== url` here, as we are checking `url`

src/blocks/image/deprecated.js
RajkiranBagal, line 1: Add deprecation in our newly decided format of files & also at the end of the object

RajkiranBagal (changes requested):

rahulvarma722 (approved):

src/blocks/info-box/block.js
```diff
@@ -34,5 +34,7 @@ registerBlockType( 'uagb/info-box', {
 	category: uagb_blocks_info.category,
 	edit: ( props ) => ( props.attributes.isPreview ? <PreviewImage image="info-box" /> : <Edit { ...props } /> ),
 	save,
+	__experimentalLabel: ( atts ) =>
+		applyFilters( 'uag_loop_data_source_label', __( 'Heading', 'ultimate-addons-for-gutenberg' ), atts ),
```
adi3890, line 38: default should be block name.
Info Box here

src/blocks/image/save.js
abhishekrijal: Actually it was to handle both case of '' and undefined. @RajkiranBagal

src/blocks/image/deprecated.js
RajkiranBagal: Check advanced heading depreciation of folder - v2.3.4

RajkiranBagal (approved):

PR #3096, opened 2023-05-08
SPCTR-2813 Resolving container margin issue 

rahulvarma722:
### Description
https://brainstormforce.atlassian.net/browse/SPCTR-2813

### Checklist:
- [ ] My code is tested


src/blocks/container/styling.js
```diff
@@ -285,8 +289,8 @@ function styling( attributes, clientId, name ) {
 		'padding-right': generateCSSUnit( rightPaddingDesktop, paddingType ),
 		'margin-top': generateCSSUnit( topMarginDesktop, marginType ) + ' !important',
 		'margin-bottom': generateCSSUnit( bottomMarginDesktop, marginType ) + ' !important',
-		'margin-left': generateCSSUnit( leftMarginDesktop, marginType ),
-		'margin-right': generateCSSUnit( rightMarginDesktop, marginType ),
+		'margin-left': generateCSSUnit( leftMarginDesktop, marginType ) + ' !important',
```
aapatre, line 292: @rahulvarma722 Is it possible to eliminate the usage of `!important` and instead increase the specificity of the selector?

Due to `!important` users cannot apply their own custom CSS (although this is editor-only, it might ruin WYSIWYG (What-You-See-Is-What-You-Get) experience.

rahulvarma722: Yes, this is possible but for it, we need to change our entire selector previously we had implemented ! important for margin-top and margin-left that's why I have implemented ! important and one more thing for this is also avoiding for theme styling.

aapatre (approved):

PR #3099, opened 2023-05-08
SPCTR-2819 - Fix: [Container] Height varies in frontend and editor and infobox alignment issue on front-end

supriyajjadhav:
### Description
[Container] Height varies in frontend and editor

### Types of changes

### Checklist:
- [x] My code is tested


src/blocks/container/styling.js
```diff
@@ -315,7 +315,6 @@ function styling( attributes, clientId, name ) {
 	}
 
 	selectors[ containerFlexSelector ] = {
-		'min-height': generateCSSUnit( minHeightDesktop, minHeightType ),
 		'flex-direction': directionDesktop,
```
adi3890, line 318: Just a concern since we are removing min-height value here.
How  will the min-height option work.
Was this added in wrong place and hence we need to remove?

supriyajjadhav: Yes, We were adding min-height to the parent selectors and `block-editor-block-list__layout` also.  Instead of adding twice we can apply min-height to the parent selector directly.
I have tested this for responsive devices as well. It is working fine.

adi3890 (approved):

PR #3109, opened 2023-05-09
Fix: Form is not submitted when we use the form block inside the footer widget.

supriyajjadhav:
### Description
Fix: Form is not submitted when we use the form block inside the footer widget.

### Types of changes

### Checklist:
- [x] My code is tested

### DIFF
--- a/blocks-config/forms/class-uagb-forms.php
+++ b/blocks-config/forms/class-uagb-forms.php
@@ -152,8 +152,21 @@ if ( ! class_exists( 'UAGB_Forms' ) ) {
 						}
 					}
 				}
-			}
+			} else {
+				$widget_content = get_option( 'widget_block' );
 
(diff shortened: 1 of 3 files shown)


AbhijitNage123 (approved):

blocks-config/forms/class-uagb-forms.php
```diff
@@ -152,8 +152,21 @@ public function process_forms() {
 						}
 					}
 				}
+			} 
+			
+			$widget_content = get_option( 'widget_block' );
+			
+			if ( ! empty( $widget_content ) && is_array( $widget_content ) && empty( $current_block_attributes ) ) {
```
adi3890, line 159: will it be possible to return early instead of nesting

adi3890 (approved):

PR #3127, opened 2023-05-11
SPCTR-2843 - Container Background Fixes

bsf-zanev:
### Description
1. Fixed the Background Generation in the Editor for Container.
2. Added Default Background for Popup Builder without Image and Updated Front-end Responsive Background.

### Types of changes

### How has this been tested?
Tested in all four cases stated in the task comment.

### Checklist:
- [x] My code is tested


src/blocks/container/backgroundCss.js
```diff
@@ -19,7 +19,7 @@ const backgroundCss = ( attributes, deviceType, clientId ) => {
 	const placeHolderImage = isLoopBuilderPresent ? { 
 		type: 'image', 
 		url: uagb_blocks_info.uagb_url + '/admin/assets/images/uag-placeholder.svg' 
-	} : {};
+	} : '';
```
adi3890, line 22: @rahulvarma722 @abhishekrijal can you please confirm this change. Seems like this was made for dynamic content compatibility

supriyajjadhav (approved):

src/blocks/container/backgroundCss.js
bsf-zanev: @rahulvarma722 - Please verify if we need the placeHolderImage to be an object at all times, and if so why.

src/blocks/container/backgroundCss.js
```diff
@@ -110,27 +112,29 @@ const backgroundCss = ( attributes, deviceType, clientId ) => {
 		gradientType,
 		gradientAngle,
 		selectGradient,
-		backgroundRepeat: backgroundRepeatTablet,
-		backgroundPosition: backgroundPositionTablet,
-		backgroundSize: backgroundSizeTablet,
-		backgroundAttachment: backgroundAttachmentTablet,
-		backgroundCustomSize: backgroundCustomSizeTablet,
+		backgroundRepeat: undefined !== backgroundRepeatTablet ? backgroundRepeatTablet : backgroundAttributesDesktop.backgroundRepeat,
```
AbhijitNage123, line 115: Can we write condition like - `backgroundRepeat: backgroundRepeatTablet ? backgroundRepeatTablet : backgroundAttributesDesktop.backgroundRepeat,`

bsf-zanev: Yes, this is possible - making this change across all the non-numeric values.

bsf-zanev: Cleaned and Tested all conditions in the editor.

AbhijitNage123 (approved):

src/blocks/container/backgroundCss.js
abhishekrijal: @bsf-zanev it was set as an object as the attribute that is passed to was having data type object.

However, it is not absolutely mandatory to be an object for this case, it can be empty string also.

bsf-zanev: Let's keep it as an empty string then, since the background color condition that already exists in `generateBackgroundCSS.js` is as follows:

`else if ( undefined === backgroundImage || '' === backgroundImage || 'unset' === backgroundImage )`

No use adding an empty object comparison as well, since the functionality remains the same.

RajkiranBagal (approved):

PR #3201, opened 2023-05-25
SPCTR-1542: Fix: Testimonial equal height is not as per intended.

rajmohanreddy774:
### Description

In the Testimonials block, when we enable the equal height option, it does not reflect the height change.

### Checklist:
- [ ] My code is tested


aapatre (changes requested):
There could be backwards compatibility issues here (as in the styling might differ if old users were relying on the default imgVrPAdding). Could you please confirm this with @RajkiranBagal or @supriyajjadhav ?

supriyajjadhav:
> There could be backwards compatibility issues here (as in the styling might differ if old users were relying on the default imgVrPAdding). Could you please confirm this with @RajkiranBagal or @supriyajjadhav ?

Correct, We can't modify attributes' default values directly.

aapatre (approved):

AbhijitNage123 (approved):

adi3890 (approved):

PR #3205, opened 2023-05-29
SPCTR-2874 [Core] Required enhancement in UAGMultiSelectControl regarding to query filter

adi3890:
### Description
- Adding searchable feature in UAGBMultiSelect control
- Compatible with option geoups also.

Above mentioned featured were required for Loop Builder query

### Checklist:
- [ ] My code is tested


src/components/multi-select-control/index.js
```diff
@@ -19,10 +20,11 @@ const defaultProps = {
 		label: '',
 		value: [],
 	},
+	isSearchable: false,
 	setAttributes: () => {},
 };
 
-export default function UAGMultiSelectControl( { label, options, data, setAttributes } ) {
+export default function UAGMultiSelectControl( { label, options, data, isSearchable, setAttributes } ) {
```
bsf-zanev, line 27: Since we now have more props, and chances are we'll need to add more in the future, we can do it like this.

```suggestion
export default function UAGMultiSelectControl( props ) {
	const {
		label,
		options,
		data,
		isSearchable,
		setAttributes,
	} = props
	
```

src/components/multi-select-control/index.js
```diff
@@ -48,12 +50,19 @@ export default function UAGMultiSelectControl( { label, options, data, setAttrib
 		selectedBlock
 	);
 
+	const allOptionsFlat = allOptions.reduce( ( accumulator, currentItem ) => {
+		if ( currentItem.options ) {
+			return [...accumulator, ...currentItem.options];
+		}
+		return [...accumulator, currentItem];
+	}, [] );
+
```
bsf-zanev, line 59: Will this work?

```suggestion
	const allOptionsFlat = allOptions.reduce( ( accumulator, currentItem ) => ( [
		...accumulator,
		currentItem.options ? ...currentItem.options : currentItem,
	] ), [] );

```

adi3890: thanx for suggestion 
implemented similar logic

src/components/multi-select-control/index.js
adi3890: done

mohitsbsftester:
Good to go as per testing on register block

PR #3279, opened 2023-06-16
SPCTR-2999 [Dynamic Content] Added Context as an attribute for use inside loop builder

abhishekrijal:
### Description
[Dynamic Content] Added Context as an attribute for use inside loop builder
The Attribute data is then being used as a source of postID and Post Type to fetch post specific custom meta keys inside Dynamic Content selection dropdown.

### Types of changes

### Checklist:
- [ ] My code is tested


src/blocks/advanced-heading/edit.js
```diff
@@ -34,6 +35,7 @@ const UAGBAdvancedHeading = ( props ) => {
 		// Assigning block_id in the attribute.
 		setAttributes( { block_id: clientId.substr( 0, 8 ) } );
 		setAttributes( { classMigrate: true } );
+		setAttributes( { context } );
```
rahulvarma722, line 38: Should we check the context in the attribute and based on changing context we should setAttribute context?

src/blocks/buttons-child/edit.js
```diff
@@ -35,6 +36,8 @@ const ButtonsChildComponent = ( props ) => {
 		// Assigning block_id in the attribute.
 		setAttributes( { block_id: clientId.substr( 0, 8 ) } );
 
+		setAttributes( { context } );
```
rahulvarma722, line 39: Should we check the context in the attribute and based on changing context we should setAttribute context?
Also, check in other blocks also this instance.

rahulvarma722 (changes requested):

mohitsbsftester:
Is there any backward compatibility to be checked in this PR?

src/blocks/buttons-child/edit.js
abhishekrijal: @rahulvarma722 Updated with check for context change.

src/blocks/advanced-heading/edit.js
abhishekrijal: @rahulvarma722 Updated with check for context change.

PR #3291, opened 2023-06-23
SPCTR-2786 Dynamic Content - Character Length, Helper functions for trimming captions and content

aapatre:
### Description
- Added character length feature for Dynamic Content
- Also added a helper function for trimming the content
- Also used the same helper function for Instagram caption trimming
- Managed edge cases for HTML entities

- Sister PR: https://github.com/brainstormforce/spectra-pro/pull/322
- [JIRA](https://brainstormforce.atlassian.net/browse/SPCTR-2786)

### Checklist:
- [ ] My code is tested


blocks-config/utils/Helpers.js
```diff
@@ -37,4 +37,62 @@ export function getPanelIdFromRef( ref ) {
 
 export const uagbClassNames = ( classes ) => classes.filter( Boolean ).join( ' ' );
 
-export const uagbDeepClone = ( arrayOrObject ) => JSON.parse( JSON.stringify( arrayOrObject ) );
\ No newline at end of file
+export const uagbDeepClone = ( arrayOrObject ) => JSON.parse( JSON.stringify( arrayOrObject ) );
+
+// Trim text by max character length to the last complete word. 
+export const trimTextToFullyVisibleWord = ( text, maxLength ) => {
```
adi3890, line 43: As i could see these functions are used in pro only, do we need to keep it in free.
Any specific reason?

classes/class-uagb-helper.php
```diff
@@ -1333,6 +1333,83 @@ public static function get_combined_selectors( $block_name, $selectors, $attr )
...
+
+		/**
+		 * Trim text but show only fully visible words.
+		 *
+		 * @param string  $txt The text to be trimmed.
+		 * @param integer $len Max trim length.
+		 * @param boolean $use_ellipsis Use ellipsis or not.
+		 * @return string
+		 *
+		 * @since 1.0.0
```
adi3890, line 1345: check for since.
Also please follow the docBlocker format we follow.

blocks-config/utils/Helpers.js
aapatre: Yes, as per the discussion with @bsf-zanev, we may need this as a helper for any upcoming feature/improvement that involves trimming text. Hence the reason we are keeping it in free code.

classes/class-uagb-helper.php
aapatre: Done

classes/class-uagb-helper.php
```diff
@@ -1333,6 +1333,82 @@ public static function get_combined_selectors( $block_name, $selectors, $attr )
...
+								$limited_caption,
+								0,
+								( false === $last_space_pos ) ? strlen( $limited_caption ) : min( strlen( $limited_caption ), $last_space_pos )
+							);
+						}
+					}
+				}//end if
+
+				// If the last character in the string matches any of the following, remove that character.				
+				switch ( substr( $limited_caption, -1 ) ) {
```
adi3890, line 1401: Should we use a better logic here instead of switch.
Can we implement in_array or something more appropriate.

blocks-config/utils/Helpers.js
```diff
@@ -37,4 +37,62 @@ export function getPanelIdFromRef( ref ) {
...
+			limitedCaption = limitedCaption.substr(
+				0,
+				Math.min( limitedCaption.length, limitedCaption.lastIndexOf( ' ' ) )
+			);
+		}
+	}
+
+	// Remove Trailing Characters.
+	switch ( limitedCaption.charAt( limitedCaption.length - 1 ) ) {
+		case ',':
```
adi3890, line 89: should we use includes here which looks much appropriate as per logic.

blocks-config/utils/Helpers.js
```diff
@@ -37,4 +37,62 @@ export function getPanelIdFromRef( ref ) {
...
+export const trimTextToFullyVisibleWord = ( text, maxLength ) => {
+
+	if( ! text ) {
+		return '';
+	}
+
+	let needsEllipsis = true;
+	let limitedCaption = text;
+
+	if ( needsEllipsis && text.length <= maxLength ) {
```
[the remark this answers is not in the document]
aapatre: - Nope, this was actually meant for an older logic.
- I've removed it, thanks for pointing the same.
- Also, I've made minor changes to the same now

PR #3401, opened 2023-08-09
SPCTR-3242 [security] sanitizing and escaping `$unit` variable

adi3890:
### Description
AS per a security concern shared by some researcher.
```
While reviewing a security related change made in the latest version of
your Spectra plugin, we found a minor security issue.

With the function get_css_value() in the file
/classes/class-uagb-helper.php, you are validating and escaping the
value of the variable $value, but you don't do the same with the
variable $unit. We confirmed that second variable can be user input, so
it should be secured as well. We didn't see a way that could be
exploited, so we didn't find a vulnerability caused by that.
```

Reference - https://brainstormforce.slack.com/archives/CAQ2TSA6A/p1691556015222879

We are updating the logic.
### Checklist:
- [ ] My code is tested


patilvikasj (approved):

classes/class-uagb-helper.php
```diff
@@ -221,16 +221,22 @@ public static function generate_css( $selectors, $id ) {
 		 *
 		 *  get_css_value( VALUE, 'em' );
 		 *
-		 * @param string $value  CSS value.
-		 * @param string $unit  CSS unit.
+		 * @param string|int|float $value  CSS value.
+		 * @param string           $unit  CSS unit.
 		 * @since 1.13.4
 		 */
 		public static function get_css_value( $value = '', $unit = '' ) {
```
bsf-zanev, line 228: Will keeping the default value to empty string cause an issue with the int|float types?

bsf-zanev (approved):

rahulvarma722 (approved):

classes/class-uagb-helper.php
adi3890: we can have expect different type of value from other function but default value can be one only.

AbhijitNage123 (approved):

PR #3404, opened 2023-08-10
SPCTR-3215 - Spectra AI: OpenAI Key Saving 

bsf-zanev:
### Description
- Added an admin menu setting to link the user's OpenAI key.
- Renamed the Block Settings folder to Integrations in the admin core, and tested with IG to ensure it works as intended.
- Minor update in the Instagran Admin Panel for cleaner code.

### Types of changes

### How has this been tested?
Tested by saving data through the input field and refreshing the admin to see if the option works as intended.

### Checklist:
- [x] My code is tested


admin-core/assets/src/dashboard-app/pages/settings/block-settings/OpenAISettings.js
```diff
@@ -0,0 +1,161 @@
...
+	);
+
+	// Update the Label of the API Key Button. 
+	const updateAPIButtonLabel = ( type = null ) => {
+		switch ( type ) {
+			case 'saving':
+				setOpenAIKeyLabel( __( 'Saving', 'ultimate-addons-for-gutenberg' ) );
+				break;
+			case 'invalid':
+				setOpenAIKeyLabel( __( 'Inavalid Key' , 'ultimate-addons-for-gutenberg' ) );
```
adi3890, line 32: typo please check

admin-core/assets/src/store/globalDataStore.js
```diff
@@ -59,7 +59,8 @@ const initialState = {
 	],
 	spectraFSEFonts: [],
 	themeFonts: [],
-	spectraIsBlockTheme: false
+	spectraIsBlockTheme: false,
+	openAIOptions: {},
```
adi3890, line 63: @bsf-zanev as we are using prefix **spectra** should that be used here too.
will it be object or array

phpstan-baseline.neon
```diff
@@ -143,7 +143,7 @@ parameters:
 
 		-
 			message: "#^Parameter \\#1 \\$input_settings of method UagAdmin\\\\Ajax\\\\Common_Settings\\:\\:sanitize_form_inputs\\(\\) expects array, mixed given\\.$#"
-			count: 4
+			count: 5
```
adi3890, line 146: we should not be updating these values.
lets resolve the issue instead

admin-core/ajax/common-settings.php
```diff
@@ -788,4 +789,18 @@ public function insta_refresh_all_tokens() {
 		}
 		wp_send_json_error( array( 'messsage' => __( 'Failed to refresh tokens', 'ultimate-addons-for-gutenberg' ) ) );
 	}
+
+	/**
+	 * Save setting - Updates the OpenAI Key.
```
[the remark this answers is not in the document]
bsf-zanev: Updated the doc comment in [this commit](https://github.com/brainstormforce/ultimate-addons-for-gutenberg/pull/3404/commits/07af0078635b072a65eab6fd7711a93e20058779).

admin-core/assets/src/store/globalDataStore.js
bsf-zanev: Not necessarily since this was originally non-prefixed, but I've updated the same - we'll follow prefixing here from now on.

As for the datatype, both yield the same result, but empty array is the more accurate approach here. Updated the same in [this commit](https://github.com/brainstormforce/ultimate-addons-for-gutenberg/pull/3404/commits/07af0078635b072a65eab6fd7711a93e20058779).

PR #3427, opened 2023-08-20
Fixed: Form block (uagb/forms) support in block themes

Ninos:
### Description
Before it was not always possible to use `uagb/forms` in sites templates (block themes).

### Reason
`$current_block_attributes` must only be set if not null/empty.

PS: Can you please also check following MR (#3200)? It just adds an additional filter, which we rly need... <3

### DIFF
--- a/blocks-config/forms/class-uagb-forms.php
+++ b/blocks-config/forms/class-uagb-forms.php
@@ -215,7 +215,7 @@ if ( ! class_exists( 'UAGB_Forms' ) ) {
 						$template_post_content = $template->content . ( ! empty( $post_content ) ? $post_content : '' );
 						$template_content      = parse_blocks( $template_post_content );
 						if ( get_template() === $template->theme && ! empty( $template_content ) && is_array( $template_content ) ) {
-							$current_block_attributes = $this->recursive_inner_forms( $template_content, $block_id );
+							$current_block_attributes = $this->recursive_inner_forms( $template_content, $block_id ) ?: $current_block_attributes;
 						}
 					}
 				}


blocks-config/forms/class-uagb-forms.php
```diff
@@ -215,7 +216,7 @@ public function process_forms() {
 						$template_post_content = $template->content . ( ! empty( $post_content ) ? $post_content : '' );
 						$template_content      = parse_blocks( $template_post_content );
 						if ( get_template() === $template->theme && ! empty( $template_content ) && is_array( $template_content ) ) {
-							$current_block_attributes = $this->recursive_inner_forms( $template_content, $block_id );
+							$current_block_attributes = $this->recursive_inner_forms( $template_content, $block_id ) ?: $current_block_attributes;
```
AbhijitNage123, line 219: @Ninos We already defined `$current_block_attributes` to false  on Line No.188. Is there any reason we are again passing it to the else case?

AbhijitNage123:
@Ninos Please change your base branch to `dev`. and make sure it is compatible with `dev`.

blocks-config/forms/class-uagb-forms.php
Ninos: Yeah, the reason is, that there's just a check if template has content, but not if there's a valid content with same `$block_id` gotten by request. So in `$this->recursive_inner_forms( $template_content, $block_id )` you may get an empty return. We could also check the returnment and if not empty -> write into `$current_block_attributes` & break loop, but upper solution is also enough :-)

Ninos:
> @Ninos Please change your base branch to `dev`. and make sure it is compatible with `dev`.

Thx, mapped to dev, seems I don't need any rebases... :-)

blocks-config/forms/class-uagb-forms.php
AbhijitNage123: @Ninos We are already checking empty condition on line no 275 and sending JSON error msg in case if we get an empty from recursive function hence no need above condition. Please revert the changes.

Ninos: This is not working here. If you have multiple templates and last loop iteration has a block but not the specific `$block_id` it returns null -> form is not working in site templates.

Check this function:
```php
# Loop over all available templates
foreach ( $template_query_posts as $post ) {
	...
	# Just checks for $template_content is not empty & template is same
	if ( get_template() === $template->theme && ! empty( $template_content ) && is_array( $template_content ) ) {
		# Current behaviour!
		# Always overwrites $current_block_attributes, also if ->recursive_inner_forms() is null/empty
		$current_block_attributes = $this->recursive_inner_forms( $template_content, $block_id );

		# To fix that, write something like that instead:
		# If ->recursive_inner_forms() is null/empty, use old value of $current_block_attributes (empty or already found block attributes)
		$current_block_attributes = $this->recursive_inner_forms( $template_content, $block_id ) ?: $current_block_attributes;

		# This would also be possible:
		# First set new value, if not empty -> set to $current_block_attributes & break
		$new_block_attributes = $this->recursive_inner_forms( $template_content, $block_id );
		if( $new_block_attributes ) {
			$current_block_attributes = $new_block_attributes;
			break;
		}
	}
}
```

Ninos:
resolved latest conflicts

PR #3441, opened 2023-08-29
Spectra v2.7.7

adi3890:
* Improvement: Container - Added an image overlay option for the background.
* Fix: Container - Resolved an issue wherein the blocks inside Containers would appear underneath the Container's overlay when set.
* Fix: Container - Resolved an issue in which the flex properties would not work as intended in the frontend for the Image and Heading blocks in nested Containers.
* Fix: Modal - Resolved an issue in which the Modal would not open in Astra's quick-view popup for WooCommerce products.
* Fix: Post Blocks - Resolved an issue wherein taxonomies were not accessible to the Shop Manager role.
* Fix: Separator - Resolved an edge-case issue in which the responsive alignment would not work as intended.
* Fix: Table Of Contents - Resolved an issue in which Latin characters in a heading would prevent the links for that specific heading from working.
* Fix: Table Of Contents - Resolved an issue wherein the collapsible setting would not work in the editor when enabled.
* Fix: Table Of Contents - Resolved an issue wherein regenerating assets would cause the Table Of Contents to appear empty in FSE.
* Fix: Global - Resolved an issue wherein a PHP warning could occurred in a few cases on some websites.


AbhijitNage123 (approved):

src/blocks/container/style.scss
```diff
@@ -156,3 +157,13 @@ body .ast-container .wp-block-uagb-container > .uagb-container-inner-blocks-wrap
 	}
 }
 
+.uagb-container-overlay {
+	content: "";
```
bsf-zanev, line 161: Do we need the `content` property for this class?

src/blocks/container/styling.js
```diff
@@ -219,7 +220,7 @@ function styling( attributes, clientId, name, deviceType ) {
 		( ( 'color' === overlayType && backgroundVideoColor ) || ( 'gradient' === overlayType && gradientValue ) )
 			? 1 - overlayOpacity
 			: 1;
-
+
```
bsf-zanev, line 223:
```suggestion

```

src/blocks/container/styling.js
```diff
@@ -276,6 +277,9 @@ function styling( attributes, clientId, name, deviceType ) {
 		' .uagb-container__video-wrap video': {
 			'opacity': backgroundVideoOpacityValue,
 		},
+		' .uagb-container-overlay': {
```
bsf-zanev, line 280: On further diving, I've noticed that we don't use `uagb-container-overlay` in the latest version of the image overlay.

We can remove it from all instances in the code.

src/blocks/separator/styling.js
```diff
@@ -5,6 +5,30 @@
 import generateCSS from '@Controls/generateCSS';
 import generateCSSUnit from '@Controls/generateCSSUnit';
 import { getFallbackNumber } from '@Controls/getAttributeFallback';
+const alignmentCSS = ( align ) => {
```
bsf-zanev, line 8: This could be in a helper, similar to its PHP counterpart.

src/blocks/table-of-contents/render.js
```diff
@@ -34,8 +33,7 @@ const Render = ( props ) => {
 
 	useEffect( () => {
 		if ( UAGBTableOfContents ) {
-			const baseSelector = classMigrate ? '.uagb-block-' : '#uagb-toc-';
```
bsf-zanev, line 37: Will removing the class migrate have an issue for the very old users that update to this version?

src/blocks/separator/styling.js
AbhijitNage123: @bsf-zanev alignmentCSS() function Is not needed for any other block. If needful in future we can move it in helper but for now none of our block use that function except Separator. Se we keep it in Separator `styling.js`.

rahulvarma722 (approved):

src/blocks/table-of-contents/render.js
AbhijitNage123: @bsf-zanev After the V2.0.0 we are only considering v1.25.6 as very old user and classMigrate is use lastly when we move by adding class `.uagb-block-${block_id}` in class attribute instead of as ID ( `#uagb-toc-${block_id}` ) to div. So it will not create any issue.

src/blocks/container/style.scss
rajmohanreddy774: Removed in this [PR](https://github.com/brainstormforce/ultimate-addons-for-gutenberg/pull/3442).

src/blocks/container/styling.js
rajmohanreddy774: updated.

src/blocks/container/styling.js
rajmohanreddy774: Removed in this [PR](https://github.com/brainstormforce/ultimate-addons-for-gutenberg/pull/3442).

rajmohanreddy774 (approved):

bsf-zanev (approved):
Looks good to me.

PR #3444, opened 2023-08-29
SPCTR-3297: Optimization-PR for container CSS generation

rajmohanreddy774:
### Description

In frontend.css.js we are generating CSS of background and overlay separately for each device which can be optimized using a function that can generate CSS based on device type, so that we can reduce the number of lines that is redundant.

### Checklist:
- [x] My code is tested


classes/class-uagb-block-helper.php
```diff
@@ -1408,6 +1408,66 @@ public static function get_masonry_gallery_css() {
...
+		 * @param array  $attr         The array of attributes containing styling options.
+		 * @param string $device_type   The device type ('Desktop', 'Tablet', 'Mobile') for which to generate background CSS.
+		 * @param string $overlay      The overlay option ('no' or 'yes') to determine whether to include overlay CSS.
+		 *
+		 * @return array               The background CSS object for the specified device type.
+		 */
+		public static function get_background_css_by_device( $attr, $device_type, $overlay = 'no' ) {
+			$bg_obj = array(
+				'backgroundType'                  => $attr['backgroundType'],
+				'backgroundImage'                 => $attr[ "backgroundImage$device_type" ],
```
adi3890, line 1426: any specific reason to keep that extra space for "" in []
can we keep all as double quotes it does not look symetrric
or keep in single quotes concatinated with php variables.

classes/class-uagb-block-helper.php
```diff
@@ -1408,6 +1408,66 @@ public static function get_masonry_gallery_css() {
...
+		 * the corresponding background object and overlay CSS for that device type.
+		 *
+		 * @param array  $attr         The array of attributes containing styling options.
+		 * @param string $device_type   The device type ('Desktop', 'Tablet', 'Mobile') for which to generate background CSS.
+		 * @param string $overlay      The overlay option ('no' or 'yes') to determine whether to include overlay CSS.
+		 *
+		 * @return array               The background CSS object for the specified device type.
+		 */
+		public static function get_background_css_by_device( $attr, $device_type, $overlay = 'no' ) {
+			$bg_obj = array(
```
adi3890, line 1424: having a switch case for $device_type will be great to avoid any issue with typo ever.
with default to desktop

includes/blocks/container/frontend.css.php
```diff
@@ -264,8 +190,8 @@
 	),
 	$border_tablet
 );
-$container_tablet_css            = array_merge( $container_tablet_css, $container_bg_css_tablet );
-$inner_container_tablet_css      = array(
+$container_tablet_css       = array_merge( $container_tablet_css, $container_bg_css_tablet );
```
adi3890, line 193: should we check if second array is not empty and perform array_merge if required

rajmohanreddy774: Added check.

classes/class-uagb-block-helper.php
rajmohanreddy774: Added switch case to handle case-insensitive deviceType input and added default to be Desktop.

classes/class-uagb-block-helper.php
rajmohanreddy774: Updated with single quotes concatenated,

We are getting phpcs error if we don't add extra spaces.
Error: `Array keys must be surrounded by spaces unless they contain a string or an integer.`

adi3890 (approved):

classes/class-uagb-block-helper.php
```diff
@@ -1408,6 +1408,80 @@ public static function get_masonry_gallery_css() {
...
+		 * This function takes attributes for styling and a device type, and returns
+		 * the corresponding background object and overlay CSS for that device type.
+		 *
+		 * @param array  $attr         The array of attributes containing styling options.
+		 * @param string $device_type   The device type ('Desktop', 'Tablet', 'Mobile') for which to generate background CSS.
+		 * @param string $overlay      The overlay option ('no' or 'yes') to determine whether to include overlay CSS.
+		 *
+		 * @return array               The background CSS object for the specified device type.
+		 */
+		public static function get_background_css_by_device( $attr, $device_type = 'Desktop', $overlay = 'no' ) {
```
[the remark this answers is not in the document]
rajmohanreddy774: Added.

adi3890 (approved):

PR #3445, opened 2023-08-30
SPCTR-3226: Fix - Global - Post related blocks have mobile and tablet styling issues.

rajmohanreddy774:
### Description

All the Post blocks mobile and tablet stylings were not working.

This fixes all these issue 
https://brainstormforce.atlassian.net/browse/SPCTR-3226
https://brainstormforce.atlassian.net/browse/SPCTR-3187
https://brainstormforce.atlassian.net/browse/SPCTR-3181

### Checklist:
- [x] My code is tested


blocks-config/uagb-controls/generateCSSUnit.js
```diff
@@ -1,4 +1,4 @@
-function generateCSSUnit( value, unit ) {
+function generateCSSUnit( value, unit = 'px' ) {
```
rahulvarma722, line 1: This is not the correct way as this is the global function and we can not put px as default value.

rajmohanreddy774: whenever the unit is undefined it will take 'px' as default, It doesn't cause any issue if the unit is defined I have verify it

bsf-zanev: Both sides of this chat are perfectly valid.

@rahulvarma722 - The default unit should be present, since we're validating the value, but not the unit. The current code will throw an error if the function call sends bad data to the unit, or doesn't send the unit.

@rajmohanreddy774 - Although Pixels should be the default for almost every CSS property, and almost every property that needs a unit will have to have it, there are some properties that function differently if they do not have a unit.

Here's one of the most well known examples, the `line-height`.
If the `line-height` does not have a unit after the value, it multiplies the `font-size` by it's value.

```
font-size: 8px;
line-height: 12px;
```
is the same as
```
font-size: 8px;
line-height: 1.5;
```

This is just one example. To avoid any breaking changes across our plugin, we can set the default to '' if it doesn't exist. This will allow all function calls to work as intended, and if a unit was not specified no harm will be done. Doing this will ensure future-proofing as well.

bsf-zanev (commented):
Changelog update, and one minor comment.

bsf-zanev (approved):
Changes look good to me.

blocks-config/uagb-controls/generateCSSUnit.js
rajmohanreddy774: @bsf-zanev Passing default unit = '' to generateCSSUnit and added a fallback unit in generateBorderCSS.

README.md
```diff
@@ -215,6 +215,9 @@ Our external packages use [Rating Star Component](https://github.com/n49/react-s
 
 ## Changelog ##
 
+### x.x.x ###
+* Fix: Post - Resolved issues where Post related blocks resposive tablet and mobile styles not working as intented.
```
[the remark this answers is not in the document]
rajmohanreddy774: Done.

README.md
```diff
@@ -218,6 +218,7 @@ Our external packages use [Rating Star Component](https://github.com/n49/react-s
 ### x.x.x ###
 * Fix: Container - Resolved an issue in which the link tag would not work as intended on the front-end.
 * Fix: How To - Resolved an issue in which the minimum minutes setting would be unskippable.
+* Fix: Post Blocks - Resolved an issue in which the resposive styling would not work as intented in the editor.
```
bsf-zanev, line 221:
```suggestion
* Fix: Post Blocks - Resolved an issue in which the responsive styling would not work as intended in the editor.
```

rajmohanreddy774: Done.

bsf-zanev (approved):

adi3890 (approved):

PR #3536, opened 2023-10-11
Spectra v2.8.0

adi3890:
* Feature: Container - Added a new Sticky-Container setting in the Advanced tab.
* Feature: Visibility Mode - Improved the Coming-Soon Mode by adding the Maintenance Mode feature.
* Improvement: Container - Updated responsive padding settings to inherit from bigger devices by default.
* Improvement: Buttons - Added alignment options to both the child and parent blocks in the block controls toolbar.
* Improvement: Buttons - Reduced the default number of buttons to 1.
* Improvement: Form - Made the default recipient as the current user's WordPress email address when the `To` field is left empty in submitted forms.
* Improvement: Global - Enhanced the reusable block functionality for the following blocks: Tabs and Slider.
* Fix: Container - Resolved an issue wherein the left and right margins of inner-containers would not work as intended.
* Fix: Content Timeline - Resolved an issue in which the icons would appear over-sized and the content styling would break on the frontend.
* Fix: Image Gallery - Resolved an issue with the masonry type wherein the loader would not be hidden after all the images have been loaded.
* Fix: Post Carousel - Resolved an issue in which the responsive columns would not work as intended in the editor.
* Fix: Post Grid - Resolved an issue wherein the post meta responsive font-size would not work as intended in the editor.
* Fix: Separator - Resolved an issue wherein the height would revert from percentage to pixels after reloading the editor.
* Fix: Global - Removed responsive conditions for core PHP-rendered blocks.
* Fix: Global - Resolved an issue in which the font-family that was set would not appear as intended in the FSE template editor.


blocks-config/blocks-config.php
```diff
@@ -5,11 +5,11 @@
  * @package UAGB
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) || ! defined( 'UAGB_DIR' ) ) {
+	exit; // Exit if accessed directly, or if UAFB_DIR is not defined.
```
AbhijitNage123, line 9: It should be `UAGB_DIR`

rajmohanreddy774 (approved):

rahulvarma722 (approved):

blocks-config/blocks-config.php
bsf-zanev:
```suggestion
	exit; // Exit if accessed directly, or if UAGB_DIR is not defined.
```

src/blocks/extensions/advanced-positioning/attributes.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * ADVANCED SETTINGS: Position - Attributes.
+ */
+
+const attributes = {
```
AbhijitNage123, line 5: All attribute which newly introduce should be unset for PHP render block else will get error like Block can not be previewed.

AbhijitNage123: Fixed here - https://github.com/brainstormforce/ultimate-addons-for-gutenberg/pull/3538

blocks-config/blocks-config.php
AbhijitNage123: Fixed here - https://github.com/brainstormforce/ultimate-addons-for-gutenberg/pull/3538

bsf-zanev (approved):
Looks good, we just need to get #3538 merged before release.

AbhijitNage123 (approved):

PR #3580, opened 2023-11-02
SPCTR-3480: Fix - Container gradient overlay not working as intended.

rajmohanreddy774:
### Description

Container background overlay gradient color has no on change of colors when picked different colors than default .
https://brainstormforce.atlassian.net/browse/SPCTR-3480

### Screenshots

https://d.pr/v/1aRfdv

### Checklist:
- [x] My code is tested


src/components/background/index.js
```diff
@@ -922,7 +922,7 @@ const Background = ( props ) => {
 								<>
 									<div className="uag-background-image-overlay-gradient">
 										<GradientSettings
-											backgroundGradient={ props.backgroundOverlayGradient }
+											backgroundGradient={ props.backgroundGradient }
```
bsf-zanev, line 925: Is this backward compatible?

Do the users that are using `backgroundOverlayGradient` lose all their changes?

If so, we may have to handle `backgroundOverlayGradient` migration.

bsf-zanev (commented):
Had one concern.

src/components/background/index.js
rajmohanreddy774: Actually we are trying to reuse the gradient component for overlay gradient as well, but there was a mistake made that the backgroundOverlayGradient variable was used where it should be backgroundGradient so the value was not update and only the default value was being applied, we can discuss this on call for more clarity.

bsf-zanev (approved):
Looks good to me.

AbhijitNage123 (approved):

adi3890 (approved):

adi3890 (approved):

PR #3604, opened 2023-11-11
Fix siteground caching issue

adi3890:
### Description
- Resolves Site Ground caching issue
[JIRA Card](https://brainstormforce.atlassian.net/browse/SPCTR-3517)

### Screenshots
[https://d.pr/v/x74GZs](https://d.pr/v/x74GZs)

### Checklist:
- [ ] My code is tested


classes/class-uagb-caching.php
```diff
@@ -0,0 +1,82 @@
...
+
+		if ( ! isset( self::$instance ) || null === self::$instance ) {
+			self::$instance = new self();
+
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor
```
bsf-zanev, line 44: We could have a since tag on this class, and this constructor.

classes/class-uagb-caching.php
```diff
@@ -0,0 +1,82 @@
...
+	 */
+	public function __construct() {
+		add_action( 'uagb_delete_uag_asset_dir', array( $this, 'clear_cache' ) );
+		add_action( 'uagb_delete_page_assets', array( $this, 'clear_cache' ) );
```
bsf-zanev, line 48: Do we have to have these two as separate actions, or can we have a single action`uagb_delete_cached_assets` that is done in `delete_page_assets()` as well as `delete_uag_asset_dir()`?

adi3890: Keeping separate actions gives more flexibility as some other specific action can be performed.
In future if some one want to add any functionality only after delete_page assets that can be manages separately.

classes/class-uagb-caching.php
adi3890: done

bsf-zanev (approved):
Looks good to me.

PR #3611, opened 2023-11-14
SPCTR-3525 - Zip AI: Added a Complete Disable Option to the Spectra Dashboard.

bsf-zanev:
### Description

- Integrated the Zip AI Library from the branch with the same ID into Spectra.
- Created an admin settings option to Disable Zip AI Chat Sidebar in Spectra.

### Screenshots
image

### Types of changes

### How has this been tested?
Tested by loading the Updated Zip AI Library into Spectra, enabling multiple plugins with Zip AI loader, and disabling it from the Spectra Admin Dashboard while ensuring that it is completely unloaded and does not load anything else, no matter how many plugins may be trying to load the library. 

### Checklist:
- [x] My code is tested
- [x] My code follows accessibility standards 


lib/zip-ai/classes/admin-configurations.php
```diff
@@ -158,13 +157,44 @@ public function verify_zip_ai_authorization() {
...
+		// Disable the Zip AI Library.
+		Functions::disable();
+
+		// Send the status based on whether the Zip AI Library is enabled or not.
+		if ( Functions::is_enabled() ) {
+			wp_send_json_error();
+		} else {
+			wp_send_json_success();
+		}
+	}
```
adi3890, line 189: Can we do this 
```suggestion
		// Send the status based on whether the Zip AI Library is enabled or not.
		if ( ! Functions::disable() ) {
			wp_send_json_error();
		} else {
			wp_send_json_success();
		}
	}
```

lib/zip-ai/classes/admin-configurations.php
```diff
@@ -158,13 +157,44 @@ public function verify_zip_ai_authorization() {
...
+		// Verify the nonce.
+		check_ajax_referer( 'zip_ai_admin_nonce', 'nonce' );
+
+		// Check if the Zip AI Assistant was requested to be disabled.
+		if ( ! empty( $_POST['disable_zip_ai_assistant'] ) ) {
+			Functions::disable_module( 'ai_assistant' );
+		}
+
+		// Disable the Zip AI Library.
+		Functions::disable();
```
adi3890, line 181: What if $_POST['disable_zip_ai_assistant'] is empty
currently its directly disabling the library should that be the logic or library should also be disabled only when we have proper $_POST variable set

adi3890 (approved):

adi3890 (approved):

PR #3618, opened 2023-11-20
Resolved an issue of style not loading for  custom post type template

AbhijitNage123:
### Description

- https://brainstormforce.atlassian.net/browse/SPCTR-3541

### Screenshots

- https://d.pr/v/ugH3Zv

### Types of changes

### Checklist:
- [x] My code is tested


rajmohanreddy774 (approved):

classes/class-uagb-post-assets.php
```diff
@@ -319,6 +319,11 @@ private function determine_template_post_type( $post_id ) {
 			// Will return true if you are using a static page as the homepage.
 			// Run only if you are on the main website URL i.e., example.com.
 			return 'page';
+		} elseif ( is_single() && is_singular() ) { // Applies to single Posts, and existing single post of any post type (post, attachment, page, custom post types).
```
adi3890, line 322: as per the description should't we go with is_singular()

what is the benefit of above logic
`is_single() && is_singular()`

AbhijitNage123: @adi3890 I wanted make sure current page is a singular post or page (including attachments) and that will be achieve using above condition. is_single() will check any post type except attachments.

AbhijitNage123: And Yes, we can go with `is_singular()`

AbhijitNage123: Updated the same and verified. Please review

classes/class-uagb-post-assets.php
```diff
@@ -319,7 +319,7 @@ private function determine_template_post_type( $post_id ) {
 			// Will return true if you are using a static page as the homepage.
 			// Run only if you are on the main website URL i.e., example.com.
 			return 'page';
-		} elseif ( is_single() && is_singular() ) { // Applies to single Posts, and existing single post of any post type (post, attachment, page, custom post types).
+		} elseif ( is_singular() ) { // Applies to single Posts, and existing single post of any post type (post, attachment, page, custom post types).
```
adi3890, line 322: Please verify few things in this function.
- Do we need is_singular check if is_page is checked for more details please check these docs
[is_page](https://developer.wordpress.org/reference/functions/is_page/), [Ref 2](https://digwp.com/2009/10/difference-between-is_singular-and-is_single/)
- if is_page is already checked do we need that in loop below.

Please test it once for any cndition where you have any doubt and update coditions accordingly.

classes/class-uagb-post-assets.php
```diff
@@ -319,6 +319,11 @@ private function determine_template_post_type( $post_id ) {
 			// Will return true if you are using a static page as the homepage.
 			// Run only if you are on the main website URL i.e., example.com.
 			return 'page';
+		} elseif ( is_singular() ) { // Applies to single Posts, and existing single post of any post type (post, attachment, page, custom post types).
+			$object = get_queried_object();
+			if ( $object instanceof WP_Post && ! empty( $object->post_type ) ) {
+				return 'single-' . $object->post_type;
+			}
```
adi3890, line 326: Also verify for this logic
is_home() && ! is_front_page()
return home

do we need this logic.
any how if that condition is not satified we retuen home in later stage.
any specific reason for same.

AbhijitNage123: @adi3890 It seems that you posted comment wrong line. when `is_home() && ! is_front_page()` this condition satisfy it means its blog page i.e example.com/static_page_as_post_page. I think I already added a inline comment inside that case.

classes/class-uagb-post-assets.php
AbhijitNage123: @adi3890 
- Yes we need `is_singular()` for custom post type which return slug `single-custom-post-type-name` and is_page when we are on single page which return slug `page` to get template by slug name.
- `is_page` Removed from the loop. its not needed.

adi3890 (approved):

adi3890 (approved):

PR #3689, opened 2023-12-26
SPCTR-3601 [Quick action bar] insert allowed blocks in selected block

adi3890:
### Description
Adding functionality where
- When a slected block allows inner block, if the block clicked fron sidebar is allowed it gets inserted iside or else it gets after it.

### Checklist:
- [ ] My code is tested


src/blocks/extensions/quick-action-sidebar/components/blocks.js
```diff
@@ -8,12 +8,15 @@ import DraggableBlock from './draggable-block';
 
 const Blocks = () => {
     const blocks = getBlockTypes();
-	const { blockInsertionPoint, getBlockRootClientId } = useSelect( ( select ) => {
+	const { blockInsertionPoint, getSelectedBlockClientId, getSelectedBlockAllowedBlocks, getBlockRootClientId } = useSelect( ( select ) => {
 		const { index } = select( 'core/block-editor' ).getBlockInsertionPoint();
```
rahulvarma722, line 12:
```suggestion
        const blockEditor = select( 'core/block-editor' );
		const { index } = blockEditor.getBlockInsertionPoint();
```
blockEditor variable, we can use in line number 13,14,15 also

rahulvarma722 (approved):

src/blocks/extensions/quick-action-sidebar/components/blocks.js
adi3890: updated

rahulvarma722 (approved):

src/blocks/extensions/quick-action-sidebar/components/blocks.js
```diff
@@ -8,12 +8,16 @@ import DraggableBlock from './draggable-block';
 
 const Blocks = () => {
     const blocks = getBlockTypes();
-	const { blockInsertionPoint, getBlockRootClientId } = useSelect( ( select ) => {
-		const { index } = select( 'core/block-editor' ).getBlockInsertionPoint();
-		const getSelectedBlockClientId = select( 'core/block-editor' ).getSelectedBlockClientId();
-		const rootClientId = select( 'core/block-editor' ).getBlockRootClientId( getSelectedBlockClientId );
+	const { blockInsertionPoint, getSelectedBlockClientId, getSelectedBlockAllowedBlocks, getBlockRootClientId } = useSelect( ( select ) => {
+		const blockEditor = select( 'core/block-editor' );
+		const { index } = blockEditor.getBlockInsertionPoint();
```
AbhijitNage123, line 13: @adi3890 can we add `blockEditor?.getBlockInsertionPoint()` after each blockEditor to make sure it does not show any error if it get undefined in any case?

adi3890: @AbhijitNage123 as we are using select( 'core/block-editor' ) inside body we will always get these value as they are not optional. 
So we will get those value.
Further if we go deeper there may be cases where we get optional value.

AbhijitNage123 (approved):

PR #3697, opened 2023-12-29
SPCTR-3618 [Core] Fix: caching issue

adi3890:
### Description
**What is the issue?**
Since most of our users are facing cahing issue due to our asset generation mechanism.
We are tweaking the url in such a wa the no user faces any caching issue.

Initially we were generating asset urls as `uag-css-12-12345678.css`
here `12` is post id and `12345678` world be timestamp at the time of assetgeneration

most of the caching plugins geenrate a static copy of html and save the same url, which check regenerated give 404 error and broken css.

**Solution we are providing**
To resolve this now we are generating assets urls as `uag-css-12.css`
and to get updated css and prevent browser cashing we are updating the version of assets so that url looks like `uag-css-12.css?ver=12345678`

This should resolve the issue of breaking websites.

> NOTE: Even after this solution if user is using file-based caching / object caching they should purge all cache in case of 
> - edit any css in specific post
> - regenrate assets.
> To automate purging process should follow [Clearing Cache with Spectra Actions - Spectra](https://wpspectra.com/docs/clearing-cache/) 
> 
> This solution prevents preaking of css / js

### Checklist:
- [ ] My code is tested


blocks-config/forms/class-uagb-forms.php
```diff
@@ -69,6 +69,7 @@ public function reusable_block_content_on_page( $reusable_ref_id ) {
 				$content = get_post_field( 'post_content', $reusable_ref_id );
 				return parse_blocks( $content );
 			}
+			return array();
```
AbhijitNage123, line 72: @adi3890 Is there any reason we are returning an empty array()? Is it because of PHPStan?

PR #3708, opened 2024-01-03
SPCTR-3604: Twenty Seventeen Icon Size Fix

yash-tingre-10:
### Description
Icon size get huge due to CSS not getting generated in Post Page

### Screenshots
https://bsf.d.pr/i/a6LsHY

### Test Cases
https://bsf.d.pr/v/CwLBo7

### Checklist:
- [x] My code is tested


classes/class-uagb-twenty-seventeen-compatibility.php
```diff
@@ -49,23 +49,13 @@ public function generate_stylesheet() {
...
-				$post    = get_post( $post_id );
-				array_push( $all_posts, $post );
-			}
-
-			if ( ! is_object( $post_assets_obj ) ) {
-				return;
-			}
-
-			foreach ( $all_posts as $post ) {
-				$post_assets_obj->prepare_assets( $post );
```
AbhijitNage123, line 68: @yash-tingre-10 Its seems that you have removed all the code which are working previously for simple page which not set as home/post page. Is it necessary to remove that?

classes/class-uagb-twenty-seventeen-compatibility.php
```diff
@@ -49,23 +49,13 @@ public function generate_stylesheet() {
...
+			if ( is_home() ) {
+				$post_id             = get_the_ID();
+				$current_post_assets = new UAGB_Post_Assets( intval( $post_id ) );
+				
+				if ( is_object( $current_post_assets ) ) {
+					$current_post_assets->enqueue_scripts();
+				}
```
AbhijitNage123, line 58: What if I used blocks on simple post/pages? Is it work in that case. Do we need to only handle the is_home() case.

AbhijitNage123 (changes requested):

classes/class-uagb-twenty-seventeen-compatibility.php
yash-tingre-10: I tested that but did not notice any change and while debugging the value came out to but null

AbhijitNage123 (approved):

adi3890 (approved):

PR #3709, opened 2024-01-03
SPCTR-3592 Fix: Global - Resolved an issue wherein the template styles for WoCoommerce template ( such as product-search-results, product_tag, product_cat, page-cart, page-checkout .etc  ) would not load as intended in FSE themes.

AbhijitNage123:
### Description

- https://brainstormforce.atlassian.net/browse/SPCTR-3592

### Screenshots

- https://d.pr/v/lTHSRm

### Types of changes

### How has this been tested?

- Added to the card

### Checklist:
- [x] My code is tested


classes/class-uagb-post-assets.php
```diff
@@ -258,20 +258,36 @@ public function __construct( $post_id ) {
 	public function get_woocommerce_template() {
 		// Check if WooCommerce is active.
 		if ( class_exists( 'WooCommerce' ) ) {
+			// Check other WooCommerce pages.
 			if ( is_cart() ) {
```
bsf-zanev, line 262: Instead of such a big if...elseif...else condition, can we do a switch true condition here?

```
switch ( true ) {
	case is_cart():
		return 'page-cart';
	case is_checkout():
		return 'page-checkout';
	case is_product():
		...
}
```

classes/class-uagb-post-assets.php
```diff
@@ -258,20 +258,36 @@ public function __construct( $post_id ) {
...
-			} elseif ( is_product_taxonomy() ) {
-				return 'taxonomy-product_cat';
-			} elseif ( is_product_tag() ) {
-				return 'taxonomy-product_tag';
-			} elseif ( is_product_category() ) {
-				return 'taxonomy-product_cat';
+			} elseif ( is_archive() ) { // Check if the current page is an archive.
+				$object = get_queried_object();
+				// Check if it's a search on the product archive.
+				$searchCondition = is_search() && is_post_type_archive( 'product' );
```
bsf-zanev, line 271: Since we're using `product-search-results` wherever the search condition is true, let's keep it as the value of the search condition and then use the Elvis operator to return.

```suggestion
				$product_search_template = ( is_search() && is_post_type_archive( 'product' ) ) ? 'product-search-results' : '';
```

Then down the line like this:
```
return $product_search_template ?: 'taxonomy-product_attribute';
```

AbhijitNage123: @bsf-zanev Returning `$product_search_template ?: 'taxonomy-product_attribute';` like this is not allowed in WordPress PHP. Getting PHPCS error (Using short ternaries is not allowed) after implementation.

classes/class-uagb-post-assets.php
AbhijitNage123: Done

classes/class-uagb-post-assets.php
```diff
@@ -258,20 +258,44 @@ public function __construct( $post_id ) {
...
+
+						case is_shop():
+							return 'archive-product';
+
+						default:
+							return $searchCondition ? 'product-search-results' : 'archive-product';
+					}
+					break;
+
+				case is_shop():
```
bsf-zanev, line 293: We have the `is_shop()` case inside the `is_archive()` case and here as well.

Do we need to have is_shop here? Will the shop page always be an archive, or is there an edge case where that condition may not be so?

AbhijitNage123: Removed unwanted condition @bsf-zanev

bsf-zanev (approved):

README.md
```diff
@@ -228,6 +228,7 @@ Our external packages use [Rating Star Component](https://github.com/n49/react-s
 * Fix: FAQ - Resolved an issue in which the answer tab would be opened by default if the content had a paragraph or break tag.
 * Fix: Post Grid - Resolved an issue wherein the featured image border-radius would not be applied and the image would overflow.
 * Fix: Global - Resolved an issue in which the Customizer responsive buttons would not work as intended.
+* Fix: Global - Resolved an issue wherein the template styles for some WooCommerce templates would not load as intended in FSE themes.
```
adi3890, line 231: Getting added to 2.11.3.
Please move to x.x.x

AbhijitNage123: Done

adi3890 (approved):

PR #3714, opened 2024-01-05
Spectra v 2.11.3

adi3890:
* Feature: Transform to - Added transform to option from Core Media & Text to Spectra Image.
* Feature: Transform to - Added transform to option from Core Buttons to Spectra Buttons.
* Feature: Transform to - Added transform to option from Core Media & Text to Spectra Infobox.
* Feature: Transform to - Added transform to option from Core Pullquote to Spectra Blockquote and Core List to Spectra Iconlist.
* Feature: Global - Added a rename option in the list view to customize Spectra block names.
* Improvement: Core - Updated asset URL generation logic to prevent caching issues.
* Fix: Container - Resolved the background overlay glitches in the style settings.
* Fix: Icon - Resolved an issue wherein the icon would appear enlarged in the Twenty Seventeen theme.
* Fix: Info Box - Resolved an issue wherein the page would reload upon clicking on the Info Box with the call-to-action set as complete box.
* Fix: FAQ - Resolved an issue in which the answer tab would be opened by default if the content had a paragraph or break tag.
* Fix: Post Grid - Resolved an issue wherein the featured image border-radius would not be applied and the image would overflow.
* Fix: Global - Resolved an issue in which the Customizer responsive buttons would not work as intended.


yash-tingre-10 (approved):

AbhijitNage123 (approved):

AbhijitNage123:
Just one suggestion @adi3890, In many places I can see getNumber, getUnit and getUnitDimention() method, Can we make that common in Helper.js and use it wherever it needed. Above method added by @yash-tingre-10 for transform.

yash-tingre-10:
There are some modifications in each function of transform, I'll have to review it once and test it so block does not break.

bsf-zanev (approved):
Just added 1 suggestion above, not a blocker for release.

rajmohanreddy774 (approved):

PR #3716, opened 2024-01-05
SPCTR-3570 - AI Features - Added the Spectra logo to the authorization screen.

bsf-zanev:
### Description
Added the plugin param to the Zip AI middleware args.

### Types of changes

### How has this been tested?
Tested with the popup in fullscreen mode.

### Checklist:
- [x] My code is tested


classes/class-uagb-loader.php
```diff
@@ -191,6 +191,7 @@ public function load_plugin() {
 			// Add the filters for the Zip AI Library and include it.
 			add_filter( 'zip_ai_collab_product_details', array( $this, 'add_zip_ai_collab_product_details' ), 20, 1 );
 			add_filter( 'zip_ai_modules', array( $this, 'add_zip_ai_modules' ), 20, 1 );
+			add_filter( 'zip_ai_auth_middleware_args', array( $this, 'add_zip_ai_auth_middleware_args' ), 20, 1 );
```
adi3890, line 194: What if multiple plugins like spectra, starter-templates, astra connecting with zip-ai is installed will this implementation give correct plugin name from it authorization is done.

adi3890: Another thought was can we add this filter in zip-ai library directly.
and add a filter to its callback so that whichever plugin wan to pass their plugin name can just pass it using one Helper function.
Every plugin don't need to manage this code.

bsf-zanev: I've updated the same in the Zip AI PR of the same task - and made the necessary changes here.

adi3890 (approved):

PR #3742, opened 2024-01-17
Spectra v2.11.4

adi3890:
### Checklist:
- [ ] My code is tested


AbhijitNage123 (approved):

yash-tingre-10 (approved):

src/blocks/tabs/deprecated.js
```diff
@@ -9,6 +9,8 @@ import { __ } from '@wordpress/i18n';
 import { getBorderAttributes } from '@Controls/generateAttributes';
 import newAttributesV2_0_14 from './deprecated/v2_0_14/attributes';
 import newSaveV2_0_14 from './deprecated/v2_0_14/save';
+import newAttributes from './attributes';
```
adi3890, line 12: @rajmohanreddy774 can you please verify if we should create a copy of attributes also.

just in case it gets updated in future.

rajmohanreddy774: @adi3890 We haven't changes in attributes.js and it won't break even if we change the attributes in future,
So @AbhijitNage123 suggested to import instead of having a copy.

rajmohanreddy774 (approved):

rahulvarma722 (approved):

src/blocks/tabs/deprecated.js
AbhijitNage123: Right, no need to add new copy of attributes.js for 2_11_3 deprecation. We haven't made any changes in attribute. It was suggested by @bsf-zanev  previously in one of my PR so I'm following the same.

bsf-zanev: @rajmohanreddy774 - For now let's create a copy of the 2.11.3 attributes to avoid any future reworking if the backward compatibility is not correctly maintained.

Further we will document the proper details on why I've suggested to use `./attributes` later on - since we need to be very careful when using it to avoid older backward compatibilities from crashing on update.

CC - @AbhijitNage123, @adi3890

rajmohanreddy774: @bsf-zanev @adi3890 Here is the PR for the changes.
https://github.com/brainstormforce/ultimate-addons-for-gutenberg/pull/3744

